PR #48, opened 2018-05-23
fix: Demo gif path broken after folder rename

nschonni:


jamarw:
Hi @nschonni , thanks for the fix!

PR #49, opened 2018-05-23
fix: homepage attribute for folder rename

nschonni:


jamarw:
Hi @nschonni , thanks for the fix!

PR #56, opened 2018-05-31
Update snippet logic

jamarw:
1. Removed unnecessary require (os module)
2. Removed conditional platform logic for slash syntax
3. Simplified path logic by using path.parse and path.relative methods
4. Update snippet links to follow CommonMark syntax

### DIFF
--- a/docs-markdown/src/helper/utility.ts
+++ b/docs-markdown/src/helper/utility.ts
@@ -97,7 +97,6 @@ export function tableBuilder(col: number, row: number) {
 export function search(editor: vscode.TextEditor, selection: vscode.Selection, searchTerm: string, folderPath: string, fullPath?: string) {
     const dir = require("node-dir");
     const path = require("path");
-    const os = require("os");
 
     if (fullPath == null) {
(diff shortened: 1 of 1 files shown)


docs-markdown/src/helper/utility.ts
```diff
@@ -114,37 +113,24 @@ export function search(editor: vscode.TextEditor, selection: vscode.Selection, s
         for (const file in files) {
             if (files.hasOwnProperty(file)) {
                 const baseName: string = (path.parse(files[file]).base);
-                let fileName: string = files[file];
-                fileName = fileName.toLowerCase();
+                const fileName: string = files[file];
                 if (fileName.includes(searchTerm)) {
```
jeahyoun: Looks like the `toLowerCase` was removed for the search term match. On Windows, this should yield the same results, but that might not be the case on Linux or Mac (where directories and filenames are case-sensitive). Just checking to see whether this was part of the criteria.

jeahyoun: We talked about this in person, so I'm approving.

PR #61, opened 2018-06-05
Update template-controller to resolve template false template notfica…

jamarw:
…tion error

1. Use checkExtension function to check for active template extension.  Execute applyTemplate command on success/active. Command will not show in quickpick, toolbar or nav menu on failure/not installed/disabled.

### DIFF
--- a/docs-markdown/src/controllers/template-controller.ts
+++ b/docs-markdown/src/controllers/template-controller.ts
@@ -1,6 +1,7 @@
 "use strict";
 
 import * as vscode from "vscode";
+import { checkExtension, generateTimestamp } from "../helper/common";
 import { reporter } from "../telemetry/telemetry";
 
 const telemetryCommand: string = "applyTemplate";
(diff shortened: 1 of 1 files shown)


docs-markdown/src/controllers/template-controller.ts
```diff
@@ -14,6 +15,11 @@ export function applyTemplateCommand() {
 
 export function applyTemplate() {
     reporter.sendTelemetryEvent("command", { command: telemetryCommand });
-    vscode.commands.executeCommand("applyTemplate").then(
-        (err) => vscode.window.showErrorMessage("Docs-article-templates extension is not installed or disabled."));
+    const extensionName = "docsmsft.docs-article-templates";
+    const { msTimeValue } = generateTimestamp();
+    const friendlyName = "docsmsft.docs-article-templates".split(".").reverse()[0];
```
wibjorn: not that it matters too much but you could:

`const [ friendlyName ] = "docsmsft.docs-article-templates".split(".").reverse();`

jamarw: Thanks @wibjorn , noted and will add to the cleanup list.

PR #65, opened 2018-06-06
Add configuration for yaml validation schema.

jeahyoun:
See https://ceapex.visualstudio.com/Engineering/_workitems/edit/23437.

1. Add TOC schema for YAML validation.
2. Add yamlValidation configuration for the RedHat plugin to consume at startup.

Testing this requires installation of the Red Hat extension: https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml#overview

### DIFF
--- a/docs-markdown/schema/toc.schema.json
+++ b/docs-markdown/schema/toc.schema.json
@@ -2,8 +2,8 @@
 	"title": "TOC",
 	"$schema": "https://dotnet.github.io/docfx/schemas/v1.0/schema.json#",
 	"version": "1.0.0",
-	"description": "The schema for a table of contents object.",
-	"id": "https://github.com/dotnet/docfx/test/schemas/Toc.schema.json",
+	"description": "The schema for a TOC object.",
+	"id": "https://github.com/dotnet/docfx/test/schemas/toc.schema.json",
 	"type": "object",
 	"additionalProperties": false,
 	"required": [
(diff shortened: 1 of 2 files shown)


jdanyow:
should add `redhat.vscode-yaml` to:
https://github.com/Microsoft/vscode-docs-authoring/blob/bee16ebe32b6173810fd1c728a01138414805e19/docs-authoring-pack/package.json#L24-L28

jeahyoun:
@jdanyow I spoke to @adkinn about this yesterday and for the time being, we're not adding this as a dependency. Not having the Red Hat extension installed doesn't result in any detrimental behavior. I think that if we move forward after the POC, we'll have the discussion whether to have this as a true dependency or a suggested installation.

jeahyoun:
It looks like there may be changes necessary to the schema file for this, as the toc.yml files are also used for breadcrumbs. This is currently on hold.

adkinn:
@jeahyoun breadcrumbs should follow the same schema since they are run through the same processor. Did you find a difference?

jeahyoun:
@adkinn There was no difference, but the original schema that I wrote didn't take breadcrumb files into account, meaning that they never would have validated correctly. I've updated the schema after gaining incredibly valuable feedback from Xinxin, who wrote the TOC reverse engineering spec.

docs-markdown/schema/toc.schema.json
```diff
@@ -0,0 +1,97 @@
+{
+	"title": "TOC",
+	"$schema": "https://dotnet.github.io/docfx/schemas/v1.0/schema.json#",
+	"version": "1.0.0",
+	"description": "The schema for a TOC object.",
+	"id": "https://github.com/dotnet/docfx/test/schemas/toc.schema.json",
```
nschonni: It doesn't look like this file exists, but I may not understand how the validator works :wink:

PR #69, opened 2018-06-07
Update alert styles

jamarw:
1. Add properties for caution, tip and important alert types.
2. Add properties for light themes.


qinezh:
LGTM :shipit:

Issue #94, opened 2018-08-04
Link UI updates causing build errors


JimDaly:
The new Link UI updates rolled out 2 days ago cause me to have build errors like the following:

> - **Line 363: [Warning]** Illegal link: `[Behavior of specialized update operations](../special-update-operation-behavior.md#behavior-of-specialized-update-operations)` -- missing bookmark. The file developer/common-data-service/special-update-operation-behavior.md doesn't contain a bookmark named behavior-of-specialized-update-operations. 

Now that the new Link UI only provides two options:
 - Heading in this file
 - Heading in another file

The error above is due to the fact that the I have to choose a heading, and the heading I want is the H1 title of the topic rather than any sub-heading. When I choose that heading I get the error above because the OPS build system seems to not recognize that a heading is a bookmark. Frankly, that makes sense to me too.

I understand why you might remove the external link option, but you should keep the option to link to another file that will include the title of the topic and a reference to the file without any fragment.

Or am I missing some other part of these new Link UI updates?

In the meantime, I have to manually strip out the generated fragments for h1 headings.

meganbradley:
Hi Jim, there are now three top level options for links. The point of the update was to reduce the number of clicks required for any link insertion. To link to another file in a repo (H1), use **Link to file in repo**. **Link to heading** is for bookmarks, hence only two options. **Link to web page** is for external (URL) links.

![image](https://user-images.githubusercontent.com/17304189/43802557-fcc86b78-9a4a-11e8-8093-4757f6bec2ec.png)

Apologies for the confusion - we didn't mean to ship without updating docs or sending a communication. Feedback on the new wording is welcome - would it help to add some clarification in the menu?

CC
@jamarw @adkinn

meganbradley:
Also, @jamarw - we should not allow the bookmark functionality to select the H1. I'll file a DCR for this.

JimDaly:
@meganbradley Thanks. 
I see the problem now.
I rely heavily on the 'legacy' toolbar which doesn't include the **Link to file in repo** command.

I'll I see is this:
![image](https://user-images.githubusercontent.com/6353736/43804931-b360a434-9a52-11e8-825b-6d9a9030ca6a.png)


I don't use the Alt+M command, I don't care for it. I like having the toolbar. I saw the rationale provided that _it takes up too much space_ - but seriously?

But at least I now know a way to work around the problem.

- Can the 'legacy' toolbar be updated to show these new commands?

[...]

PR #95, opened 2018-08-11
Update toolbar link options and heading level support

jamarw:
1. Add new function to add external and internal link options to toolbar link command
2. Change bookmark/heading command to support h2-h6 (no more h1)

Resolves issue [94](https://github.com/Microsoft/vscode-docs-authoring/issues/94)

### DIFF
--- a/docs-markdown/src/controllers/media-controller.ts
+++ b/docs-markdown/src/controllers/media-controller.ts
@@ -16,6 +16,7 @@ export function insertLinksAndMediaCommands() {
         { command: insertLink.name, callback: insertLink },
         { command: insertImage.name, callback: insertImage },
         { command: selectLinkType.name, callback: selectLinkType },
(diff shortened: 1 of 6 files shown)


docs-markdown/src/controllers/media-controller.ts
```diff
@@ -304,6 +305,38 @@ export function selectLinkType() {
...
+    if (!isValidEditor(editor, false, "insert link")) {
+        return;
+    }
+
+    if (!isMarkdownFileCheck(editor, false)) {
+        return;
+    }
+
+    const linkTypes = ["External", "Internal", "Bookmark in this file", "Bookmark in another file"];
+    vscode.window.showQuickPick(linkTypes).then((qpSelection) => {
```
wibjorn: super nit pick, no parens around `qpSelection`

PR #104, opened 2018-10-02
Improve handling of redirect_document_id Boolean

jamarw:
1. Removed redirect_document_id hardcoded 'false' value.
1. Updated logic to check for a valid yaml header and redirect_url attribute.  If this check passes, then check for the redirect_document_id attribute.

**Expected results for redirect_document_id attribute**
If present and true: value is true
If present and false: value is false
If redirect_document_id attribute is not present: value is false
If redirect_document_id attribute is present but value is empty: value is false

### DIFF
--- a/docs-markdown/src/controllers/master-redirect-controller.ts
+++ b/docs-markdown/src/controllers/master-redirect-controller.ts
@@ -40,12 +40,13 @@ export class RedirectionFile {
     // Members mapping to JSON elements in master redirection file
     public source_path: string;
     public redirect_url: string;
-    public redirect_document_id: boolean = false;
+    public redirect_document_id: boolean;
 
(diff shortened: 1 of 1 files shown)


docs-markdown/src/controllers/master-redirect-controller.ts
```diff
@@ -40,12 +40,13 @@ export class RedirectionFile {
     // Members mapping to JSON elements in master redirection file
     public source_path: string;
     public redirect_url: string;
-    public redirect_document_id: boolean = false;
+    public redirect_document_id: boolean;
```
wibjorn: Just a suggestion because the below works, but couldn't we initialize this as something? The null check on 126 could just check for something else. Anyway, going to approve, but leave this for rumination.

Issue #101, opened 2018-09-20
[Markdocs Server Error]: System.ArgumentNullException: Value cannot be null.


mmacy:
After updating VS Code, I typically see some version of the following exception upon first launch after update. Here's the latest after my update to VS Code v1.72.2 today:

```
[Markdocs Server Error]: System.ArgumentNullException: Value cannot be null.
   at System.Threading.Monitor.ReliableEnter(Object obj, Boolean& lockTaken)
   at System.Net.HttpEndPointListener.ProcessAccept(SocketAsyncEventArgs args)
   at System.Net.HttpEndPointListener.Accept(SocketAsyncEventArgs e)
   at System.Net.HttpEndPointListener..ctor(HttpListener listener, IPAddress addr, Int32 port, Boolean secure)
   at System.Net.HttpEndPointManager.GetEPListener(String host, Int32 port, HttpListener listener, Boolean secure)
   at System.Net.HttpEndPointManager.AddPrefixInternal(String p, HttpListener listener)
   at System.Net.HttpEndPointManager.AddListener(HttpListener listener)
   at System.Net.HttpListener.Start()
   at MarkdocsService.MarkdownHttpServer.Start() in C:\projects\markdocs-server\MarkdownHttpServer.cs:line 32
   at MarkdocsService.Program.HandleRequest() in C:\projects\markdocs-server\Program.cs:line 26
   at MarkdocsService.Program.Main(String[] args) in C:\projects\markdocs-server\Program.cs:line 14
```

The output window displays the following prior to the Markdocs server error popups:

```
Installing docs preview extension's dependencies...
Platform: linux, x86_64, name=ubuntu, version=18.04

Downloading package "docs-previewService for Linux (x64)" (26101 KB) .................... Done!

Installing package "docs-previewService for Linux (x64)"

Finished
```

## Environment

**OS:** Ubuntu 18.04 LTS
**VSCode version:** 1.72.2
**Docs extension versions**:
![image](https://user-images.githubusercontent.com/13394541/45827546-7ef1b780-bcab-11e8-8ae7-a6120c650e0d.png)

jamarw:
Hi @mmacy , looping in @qinezh for assistance.  Thanks

qinezh:
@mmacy Can you please try the latest version of docs-preview and see whether the issue still exists?

mmacy:
Hi @qinezh,

[...]

PR #109, opened 2018-10-17
Integrate xref service to resolve xref syntax

qinezh:
This is the long term solution for xref service preview behavior.


qinezh:
#101

PR #113, opened 2018-11-06
Custom rule updates

jamarw:
1. Separated zone-moniker rule into through different rules (syntax, zone and moniker).  Each extension will have it's own rule file going forward.  Updated rule names to follow "docs.{rule}" convention.
2. Added common.js to centralize regex and shared functions. 
3. Added checkMarkdownlintCustomProperty() to extension.ts.  This method is used to check for and preserve any markdownlint custom rules that a user has in addition to adding an entry for the docs custom rule path.
4. Deleted zone-moniker.js as it's no longer needed.


docs-markdown/markdownlint-custom-rules/common.js
```diff
@@ -0,0 +1,25 @@
+// @ts-check
+
+"use strict";
+
+// Triple colon
+module.exports.singleColon = /^:/gm;
+module.exports.tripleColonSyntax = /^:::/gm;
+module.exports.validTripleColon = /^:::\s+/gm;
```
nschonni: Does space at the beginning of the `:::` make it invalid?

jamarw: Hi @nschonni , currently the ::: should start at the beginning of a line with no whitespace.  Not sure if that may change going forward.  Thanks

nschonni: 👍 should there be a test for the the triple colon with leading spaces as a failure then?

jamarw: Good call, I'll let the PM know for follow-up.  Thanks

PR #119, opened 2018-11-28
Fix event-stream vulnerability

jamarw:
Update vscode version (package.json) and additional dependencies (package.lock).


nschonni:
:+1: I guess because it was a devDependency it didn't get yanked from the marketplace :wink:

PR #123, opened 2018-12-05
Adding a custom rule to require all links to Microsoft properties to be secure (https vs http)

DuncanmaMSFT:
First PR, feel free to rip it apart :)


docs-markdown/markdownlint-custom-rules/securelinks.js
```diff
@@ -0,0 +1,49 @@
...
+      return token.type === "inline";
+    }).forEach(function forToken(inline) {
+      inline.children.filter(function filterChild(child) {
+        return child.type === "link_open";
+      }).forEach(function forToken(link) {
+        if (link.attrs[0] && link.attrs[0][0] == "href") {
+          const linkPattern = /^(http:\/\/www\.|http:\/\/)([^\/]*)\/.*$/;
+
+          const href = link.attrs[0][1];
+          if (href.startsWith("http://")) {
```
jamarw: I don't think it's a big deal but could case be an issue?  For example, [Go get Visual Studio](HtTps://www.visualstudio.com/foo/test.txt) does not get picked up for comparison.

docs-markdown/markdownlint-custom-rules/securelinks.js
```diff
@@ -0,0 +1,49 @@
+// @ts-check
+
+"use strict";
+
+const common = require("./common");
+const detailStrings = require("./strings");
+
+module.exports = {
+  "names": ["secure-links"],
```
jamarw: We've used **docsmd.{rule}** as our naming convention to identify the rule as being docs-specific.

docs-markdown/markdownlint-custom-rules/securelinks.js
```diff
@@ -0,0 +1,49 @@
+// @ts-check
+
+"use strict";
+
+const common = require("./common");
```
jamarw: We can drop common and detailStrings since they are not used in this rule.

.vscode/launch.json
```diff
@@ -0,0 +1,17 @@
+{
```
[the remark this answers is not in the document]
DuncanmaMSFT: Removed, thanks!

docs-markdown/markdownlint-custom-rules/securelinks.js
DuncanmaMSFT: added an appropriate test and *yes* it was a problem, updated code to handle

docs-markdown/markdownlint-custom-rules/securelinks.js
DuncanmaMSFT: Fixed!

docs-markdown/markdownlint-custom-rules/securelinks.js
DuncanmaMSFT: Lets not close this yet, I think I should move the href regex to common *and* consider if I should move any error text to strings file

DuncanmaMSFT:
@jamarw I think I've resolved all the comments, and pulled in the latest changes!

PR #142, opened 2019-02-22
Learn scaffolding

jamarw:
Updated template extension to support learn scaffolding.  Users will be able to create modules and units using the existing templates command.

1. Added additional default setting values.
2. Refactored showTemplates functions: broke into two functions (displayTemplates and applyDocsTemplate), no longer creates a new empty file prior to showing options, improved sorting.
3. Updated import statements to use explicit/named exports.
4. Began pulling ux strings into strings.ts file for testing and easier updates.
5. Additional helpers for unit and module creation.


docs-article-templates/src/controllers/quick-pick-controller.ts
```diff
@@ -68,3 +86,57 @@ export async function showTemplates() {
...
+        addModuleFiles(modulePath, module);
+    } catch (error) {
+        console.log(`Function: createModuleDirectory ${error}`);
+    }
+
+}
+
+export function addModuleFiles(modulePath: string, module: string) {
+    try {
+        const introductionSrc = "E:\\GitHub\\vscode-docs-authoring\\docs-article-templates\\src\\learn-templates\\1-introduction.yml";
```
adunndevster: This looks specific to your environment? Will this be a problem?

adunndevster: Oh... I hadn't noticed this is outdated... nevermind!

PR #148, opened 2019-03-12
64232 cleanup scripts

bharney-msft:
Cleanup... is accessible via the markdown extension (Alt + M) and allows the user to update metadata for each file in the working root directory/ repo. Cleanup... does 4 main things:
- [x] Update .md and .yml files that contain metadata of Array's with a single item, and changes the syntax to use a single item value.
- [x] Update Microsoft links to use https://, and remove language specific links (en-us etc...)
- [x] Lowercases the metadata in .md files for certain metadata attributes
- [x] Allows the user to run all of these scripts at once.

The scripts also touch each file in the repo and replace values. So the total time the scripts take to run/complete will vary based on the size of the repo.  Any where from a few seconds to 4 minutes for the sql-docs and azure repos.


docs-markdown/src/controllers/cleanup-controller.ts
```diff
@@ -0,0 +1,427 @@
...
+    });
+    generateMasterRedirectionFile();
+}
+
+/**
+ * Searches through all directories from rootPath
+ * and cleans up Yaml Metadata values that have single array items
+ * then converts the array to single item.
+ */
+function handleSingValuedMetadata() {
```
nschonni: Sing ->Single

docs-markdown/src/controllers/cleanup-controller.ts
```diff
@@ -344,10 +371,38 @@ function capaitalizationOfMetadata() {
...
+    if (metadata) {
+        let captureParts = regex.exec(data);
+        let value = ''
+        if (captureParts && captureParts.length > 2) {
+            value = captureParts[2].toLowerCase();
+            try {
+                return data.replace(regex, `${variable}: ${value}`)
+            }
+            catch (error) {
+                console.error('Error occurred:', error);
```
adunndevster: I'm pretty sure it does, but when the error is logged, does it also log the line number?

docs-markdown/src/controllers/cleanup-controller.ts
```diff
@@ -0,0 +1,308 @@
...
+        label: "Capitalization of metadata values",
+    });
+
+    window.showQuickPick(items).then((selection) => {
+        if (!selection) {
+            return;
+        }
+
+        switch (selection.label.toLowerCase()) {
+            case "single-valued metadata":
```
adunndevster: Are we using typescript 2? If so we could use the readonly modifier and start moving strings like this to immutable constants.

docs-markdown/src/controllers/cleanup-controller.ts
```diff
@@ -0,0 +1,427 @@
...
+    data = data.replace(docsRegex, "https://docs.microsoft.com")
+    let azureRegex = new RegExp(/http:\/\/azure.microsoft.com/g)
+    data = data.replace(azureRegex, "https://azure.microsoft.com")
+    let msdnRegex = new RegExp(/http:\/\/msdn.microsoft.com/g)
+    data = data.replace(msdnRegex, "https://msdn.microsoft.com")
+    let technetRegex = new RegExp(/http:\/\/technet.microsoft.com/g)
+    data = data.replace(technetRegex, "https://technet.microsoft.com")
+    let docsRegexLang = new RegExp(/https:\/\/docs.microsoft.com\/[A-Za-z]{2}-[A-Za-z]{2}\//g)
+    data = data.replace(docsRegexLang, "https://docs.microsoft.com/")
+    let azureRegexLang = new RegExp(/https:\/\/azure.microsoft.com\/[A-Za-z]{2}-[A-Za-z]{2}\//g)
```
nschonni: The one time i've had to "undo" this is when it was embedded in sample JSON response from the translation API. Probably not worth coding around that though

docs-markdown/package.json
```diff
@@ -214,6 +219,11 @@
           "command": "applyTemplate",
           "group": "Docs",
           "when": "vscode.extensions.getExtension('docsmsft.docs-article-templates').active"
+        },
+        {
+          "command": "applyCleanup",
+          "group": "Docs",
+          "when": "vscode.extensions.getExtension('docsmsft.docs-article-templates').active"
```
jamarw: I think we can change this to "resourceLangId == markdown" so it doesn't have a dependency on the templates extension.

docs-markdown/src/controllers/quick-pick-menu-controller.ts
```diff
@@ -97,6 +98,10 @@ export function markdownQuickPick() {
             description: "",
             label: "$(diff) Template",
         });
+        items.push({
+            description: "",
+            label: "$(tasklist) Cleanup...",
+        });
```
jamarw: This can be moved out of the template extension check into the core quickpick menu.

bharney-msft:
> Very neat! Are some of these going to be made available through a "Format on Save" type option?

Thank you for the suggestion and that is an excellent Idea. I will see if we can make that happen. It would be far less intrusive to keep a markdown linter watching vs running a long process to fix everything all at once.

nschonni:
@bharney-msft do you want me to file a separate issue?

bharney-msft:
@nschonni Sounds good, please do file another issue, I like that idea

PR #153, opened 2019-03-13
66973 migrate docs yaml dap

adunndevster:


docs-authoring-pack/package.json
```diff
@@ -26,7 +26,11 @@
...
     "docsmsft.docs-markdown",
     "docsmsft.docs-article-templates",
-    "streetsidesoftware.code-spell-checker"
+    "streetsidesoftware.code-spell-checker",
+    "docsmsft.docs-yaml"
   ],
-  "license": "MIT"
+  "license": "MIT",
+  "dependencies": {
+    "vsce": "^1.58.0"
```
bharney-msft: Do we need this as a dependency? I though this was a global install

adunndevster: Yeah, removed :)

docs-markdown/package.json
```diff
@@ -311,6 +311,7 @@
     "test-core": "node ./node_modules/vscode/bin/test"
   },
   "dependencies": {
+    "@types/jest": "^24.0.11",
```
bharney-msft: This should be a dev dependency with the other types

docs-yaml/src/yaml-support/yaml-constant.ts
```diff
@@ -0,0 +1,16 @@
+import * as path from 'path';
```
jamarw: No big deal but we can change these to explicit imports if we follow the same style as the other extensions.  We can also enable TSLint to clean up some of the quotation/template string differences.

adunndevster: Good idea!

docs-markdown/package.json
adunndevster: Good catch.

docs-yaml/src/yaml-support/yaml-constant.ts
adunndevster: @jamarw - Updated!

docs-markdown/package.json
adunndevster: @bharney-msft - updated!

PR #154, opened 2019-03-14
Add page-level metadata to units and modules

jamarw:
1. Add page-level metadata to units and modules
2. Use **gitHubID** and **alias** settings to populate author and ms.author values if available. If a user hasn't set values, use default placeholder (this will result in build warnings as a reminder to update values).
3. Add content property to units and prerequisites to modules.


jamarw:
1. Add page-level metadata to units and modules
2. Use **gitHubID** and **alias** settings to populate author and ms.author values if available.  If a user hasn't set values, use default placeholder (this will result in build warnings as a reminder to update values).
3. Add content property to units and prerequisites to modules.

docs-article-templates/src/helper/unit-builder.ts
```diff
@@ -64,17 +84,20 @@ export async function cleanupUnit(generatedUnit: string) {
     try {
         const moduleContent = readFileSync(generatedUnit, "utf8");
         const updatedModule = moduleContent.replace("header: ", "")
-            .replace(/'/g, "");
+            .replace(/'/g, "")
+            .replace(`content: |-`, "content: |")
+            .replace(/^\s*[\r\n]/gm, "");
         writeFileSync(generatedUnit, updatedModule, "utf8");
```
bharney-msft: Is it possible to use `writeFile` or does it need to be `writeFileSync`?

docs-article-templates/src/helper/unit-builder.ts
```diff
@@ -41,16 +43,34 @@ export async function createUnits() {
...
     }
+    if (!gitHubID) {
+        author = `...`;
+    }
+    if (!alias) {
+        msAuthor = `...`;
+    }
+
     /* tslint:disable:object-literal-sort-keys */
     const yaml = require("write-yaml");
```
bharney-msft: Have you tried to use `js-yaml` Im pretty sure `write-yml` uses `js-yaml.dump(obj)` under the hood. Its a more widely used npm package.

jamarw: Hi @bharney-msft , good idea.  Updated.

PR #156, opened 2019-03-14
Removed Spell Checker from list of depedencies and added as recommend…

bharney-msft:
…ation.

### DIFF
--- a/docs-authoring-pack/.vscode/extensions.json
+++ b/docs-authoring-pack/.vscode/extensions.json
@@ -1,0 +1,10 @@
+{
+	// See http://go.microsoft.com/fwlink/?LinkId=827846 to learn about workspace recommendations.
+	// Extension identifier format: ${publisher}.${name}. Example: vscode.csharp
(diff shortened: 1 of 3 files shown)


nschonni:
Won't this just recommend it for people editing this workspace not other repos?

nschonni:
Is there a reason you're dropping this? It's found quite a few docs and product issues in the docs

bharney-msft:
> Is there a reason you're dropping this? It's found quite a few docs and product issues in the docs

I believe the rational was because the spellchecker could not be disabled when it is set as dependency. And had been displaying linting warnings in code, but no way to turn the warnings off.

nschonni:
It can be disabled for a project/workspace by adding `"cSpell.enabled": false` in the settings.json.
That is what the menu item `Disable Spell Checking For Workspace" does, although I recognize that isn't the same as disabling the plug-in fully

nschonni:
Looks like the `"cSpell.enabled": false` works at the global VS Code settings.json to turn it off for all projects

PR #164, opened 2019-03-26
Refactor Cleanup

bharney-msft:
Changed the way Cleanup scripts recursively gets the files from the nested folder directories. Now we are using a npm package that handles recursively getting a list of files from nested directories.
- Added progress indicated that displays the percent completed
- Displays completed when script is done running.
- using recursive-readdir npm package to traverse directories.
- uses promise.all to allow using async methods


docs-markdown/src/controllers/cleanup-controller.ts
```diff
@@ -3,10 +3,10 @@
 import * as vscode from "vscode";
 import { reporter } from "../telemetry/telemetry";
 import { window, workspace, ProgressLocation } from "vscode";
-import { readFileSync, writeFileSync, readdir, stat } from "fs";
-import { resolve } from "path";
+import { readFileSync, writeFileSync } from "fs";
 import { generateMasterRedirectionFile } from "./master-redirect-controller";
 import { postError, showStatusMessage } from "../helper/common";
+var recursive = require("recursive-readdir");
```
jamarw: Could this be changed to **const** for consistency and TSLint?

docs-markdown/src/controllers/master-redirect-controller.ts
```diff
@@ -193,9 +193,15 @@ export function generateMasterRedirectionFile() {
                         if (fs.existsSync(docsRedirectDirectory)) {
                             fs.mkdirSync(deletedRedirectsPath);
                         } else {
-                            fs.mkdirSync(docsAuthoringHomeDirectory);
-                            fs.mkdirSync(docsRedirectDirectory);
-                            fs.mkdirSync(deletedRedirectsPath);
+                            if (!fs.existsSync(docsAuthoringHomeDirectory)) {
```
bharney-msft: @jamarw this was getting stuck for me because I already had the directory from the templates. You might want to check that this doesn't get hung up if the directory already exists when trying to use the templates.

docs-markdown/package.json
```diff
@@ -1,343 +1,348 @@
...
+    "vscode:prepublish": "tsc -p ./",
+    "compile": "tsc -watch -p ./",
+    "postinstall": "node ./node_modules/vscode/bin/install",
+    "test": "node ./node_modules/vscode/bin/test && jest tests ./test/markdownlint-custom-rules",
+    "test-rules": "jest tests ./test/markdownlint-custom-rules",
+    "test-core": "node ./node_modules/vscode/bin/test"
+  },
+  "dependencies": {
+    "@types/jest": "^24.0.11",
+    "diff": "^4.0.1",
```
bharney-msft: Bad diff on package.json. Only change made was to add diff npm package

docs-markdown/src/controllers/cleanup-controller.ts
```diff
@@ -1,427 +1,604 @@
...
+        [".git", ".github", ".vscode", ".vs", "node_module"],
+        (err: any, files: string[]) => {
+            if (err) {
+                postError(err);
+            }
+            let percentComplete = 0;
+            let promises: Promise<{} | void>[] = [];
+            files.map((file, index) => {
+                if (file.endsWith(".yml") || file.endsWith("docfx.json")) {
+                    promises.push(new Promise((resolve, reject) => {
```
bharney-msft: Updated methods use promises and callbacks so all of the code is async.

PR #182, opened 2019-04-05
Config free docs-yaml schema loader

jamarw:
1. Create new function in extension.ts to pull json data from url.
2. Create new output channel for logging.
3. Remove SCHEMA_CONFIG_FILE constant.
4. Update loadSchema function to use mapping file data.
5. Update loadJson function to parse dynamic json data instead of using a file stored on-disk.
6. Update changelog and package.json
7. Delete local schema_config.json.


docs-yaml/src/extension.ts
```diff
@@ -60,6 +62,12 @@ async function removeTocSchemaFromConfigAtScope(value: string, scope: Configurat
     await workspace.getConfiguration().update(YAML_SCHEMA_CONFIG_NAME_OF_VSCODE_YAML_EXTENSION, newValue, scope);
 }
 
+// retrieve mapping file data and store the data in variable
+export async function loadSchemaConfig() {
+    const getResult = await WebRequest.get('https://static.docs.com/ui/latest/schemas/schema_config.json');
```
lamebrain: Is there a config file for the project that we can pull constants from?

jamarw: There is a constant file with static values that we can use: https://github.com/Microsoft/vscode-docs-authoring/blob/develop/docs-yaml/src/yaml-support/yaml-constant.ts

Do you think we should store the mappingFile constant here?

lamebrain: yeah, that looks like a good spot.

PR #196, opened 2019-04-22
Release new preview

jamarw:
1. Merging into master from develop.


docs-preview/package.json
```diff
@@ -66,89 +62,10 @@
...
-      "binaries": [
-        "./MarkdocsService"
-      ],
-      "installTestPath": "./.markdocs/MarkdocsService"
-    }
-  ]
+    "chai": "^4.2.0",
+    "markdown-it": "^8.4.2",
+    "markdown-it-include": "^1.1.0",
+    "tslint": "^5.16.0",
```
bharney-msft: I think tslint should be a dev dependency?

jamarw: Good call

docs-preview/package.json
```diff
@@ -62,93 +58,14 @@
...
-        "x86_64"
-      ],
-      "binaries": [
-        "./MarkdocsService"
-      ],
-      "installTestPath": "./.markdocs/MarkdocsService"
-    }
-  ]
-}
+    "chai": "^4.2.0",
```
nschonni: This is probably also a devDependency

jamarw: Good call

PR #198, opened 2019-04-23
Add telemetry to docs-markdown

jamarw:
1. Add telemetry events
2. Small tslint fixes

### DIFF
--- a/docs-markdown/src/controllers/alert-controller.ts
+++ b/docs-markdown/src/controllers/alert-controller.ts
@@ -48,19 +48,19 @@ export function insertAlert() {
             if (editor) {
                 insertContentToEditor(editor, insertAlert.name, formattedText, true);
                 if (qpSelection.startsWith("Note")) {
-                    reporter.sendTelemetryEvent(`${telemetryCommand}.note`, undefined, undefined);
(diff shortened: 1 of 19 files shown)


docs-markdown/src/controllers/alert-controller.ts
```diff
@@ -47,21 +47,20 @@ export function insertAlert() {
             }
             if (editor) {
                 insertContentToEditor(editor, insertAlert.name, formattedText, true);
-
                 if (qpSelection.startsWith("Note")) {
-                    reporter.sendTelemetryEvent("command", { command: telemetryCommand + ".note" });
+                    reporter.sendTelemetryEvent(`${telemetryCommand}.note`, undefined, undefined);
```
bharney-msft: Do we need to stub out the undefined? Arn't these two optional parameters?

jamarw: Good point, I think they were required in the past but probably not any longer.  I'll check it out now.

jamarw: @bharney-msft Looks much better, good idea.

PR #204, opened 2019-04-30
Jest refactoring and code coverage

jamarw:
1. Add code coverage directory to gitignore
2. Removed chai and mocha dev dependencies
3. Refactor controllers and helpers to help manage testable code

To generate code coverage report, run `npm test --coverage`


docs-markdown/package.json
```diff
@@ -353,19 +347,24 @@
     ],
     "moduleFileExtensions": [
       "ts",
-      "js"
+      "tsx",
```
bharney-msft: I dont believe we need this file extension we should be able to remove this. jsx and tsx file extensions are for react files.

jamarw: Good catch, missed those.

Issue #200, opened 2019-04-25
Preview is not displaying DocFX specific  Notes


cblackuk:
Related to the preview issue -> https://github.com/dotnet/docfx/issues/4406

Preview on old extension would render correctly like this:
![image](https://user-images.githubusercontent.com/28607803/56720081-e1c39980-6739-11e9-9613-57de8d8741d1.png)

But in this one it does that:
![image](https://user-images.githubusercontent.com/28607803/56720141-06b80c80-673a-11e9-8f04-6b8eb887f074.png)

Any chance it could be fixed? What is the point of having docs extension preview that does not render docs correctly?

jamarw:
Hi @cblackuk , we are aware of the issue and we are working on a fix.  Thanks

cblackuk:
Hi @jamarw, thanks for getting back to me. It would be great to get it fixed as we use it every single day and the old extension was very helpful in making sure DocFX formatting is correct.

jamarw:
Hi @cblackuk , just a quick follow-up on this.  There is a PR open with additional updates to the preview extension that includes alert support so look for an update next week.  Thanks

cblackuk:
Hi @jamarw  Thanks a lot! Cannot wait for that and I really appreciate you letting me know about it too.

jamarw:
No problem @cblackuk , the update has been released.  Thanks

cblackuk:
OMG it works!!! Thank you so much! @jamarw 
![image](https://user-images.githubusercontent.com/28607803/57277047-284eb780-709b-11e9-9b8f-8a2d81df497b.png)

PR #206, opened 2019-05-03
Preview alerts

jamarw:
1. Added alert-styles.css to style blockquotes.
2. Added alert-styles.js to set alert classes and structure content.
3. Remove unused css and js files.
4. Update package.json to use custom css and js.

https://github.com/Microsoft/vscode-docs-authoring/issues/200


docs-preview/media/alert-styles.css
```diff
@@ -1,16 +1,74 @@
...
-	background-position: top left;
-	background-repeat: no-repeat;
-	text-indent: 23px;
-	}
-	blockquote.NOTE span {
-		  display: block;
-		  background-repeat: no-repeat;
-		  background-position: bottom right;
-		}
+  font: 14px/20px italic Times, serif;
```
adunndevster: It looks like the only part of these blockquote styles that are not shared are the background attributes?

Maybe you can make a common blockquote.alert class with the shared values, and then keep the other items separate?

docs-preview/media/alert-styles.css
```diff
@@ -1,16 +1,74 @@
...
+blockquote.TIP {
+  font: 14px/20px italic Times, serif;
+  padding: 8px;
+  background: rgba(0, 100, 73, 0.89);
+  margin: 5px;
+  background-position: top left;
+  background-repeat: no-repeat;
+  text-indent: 23px;
+}
+blockquote.TIP span {
```
adunndevster: also all of the spans look the same, so maybe you could just assign them to blockquote.alert span?

docs-preview/media/alert-styles.js
```diff
@@ -1,19 +1,19 @@
-var blockquoteTags = document.getElementsByTagName('blockquote');
+var blockquoteTags = document.getElementsByTagName("blockquote");
```
adunndevster: Just an idea... but maybe you could make an array filled with strings of the different kinds of alerts, and then map a function onto the array to check for each kind, and assign its corresponding attribute?

Let me try working on an example...

adunndevster: @jamarw  - what about something like this?

`var blockquoteTags = document.getElementsByTagName("blockquote");
var alerts = ["NOTE", "TIP", "IMPORTANT", "CAUTION", "WARNING"];

alerts.map(aa => {
  for (var i = 0; i < blockquoteTags.length; i++) {
    let tag = blockquoteTags[i];
    if (tag.innerHTML.toUpperCase().indexOf(`[!${aa}]`) > -1) {
      tag.setAttribute("class", aa);
    }
  }
});`

jamarw: @adunndevster Great idea!  I looked at a few options similar to this and didn't come up with anything as clean.  Thx

PR #213, opened 2019-05-13
Update fenced code UI

jamarw:
1. Remove "none" as an option
2. Add following text as placeholder: "Select a programming language (required)"

### DIFF
--- a/docs-markdown/changelog.md
+++ b/docs-markdown/changelog.md
@@ -1,5 +1,10 @@
 # Change Log
 
+## 0.2.6 (TBD)
+
+- Build reporting link
+- UI updates
+
 ## 0.2.5 (April 26th, 2019)
 
 - Telemetry: basic usage data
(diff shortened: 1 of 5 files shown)


nschonni:
Looks like the constants might need to be updated before you turn this on https://docs.microsoft.com/en-us/contribute/how-to-write-use-markdown#code-snippets
The constant list is a little smaller
https://github.com/microsoft/vscode-docs-authoring/blob/dd51c6de8ce1a86b78b128ca301aaaab4a79b672/docs-markdown/src/constants/docs-code-languages.ts#L4-L23

jamarw:
> Looks like the constants might need to be updated before you turn this on https://docs.microsoft.com/en-us/contribute/how-to-write-use-markdown#code-snippets
> The constant list is a little smaller
> https://github.com/microsoft/vscode-docs-authoring/blob/dd51c6de8ce1a86b78b128ca301aaaab4a79b672/docs-markdown/src/constants/docs-code-languages.ts#L4-L23

Hi @nschonni , good catch.  We're actually handling this in a separate PR as there may be a couple tweaks to that list.  This PR will handle some validation requirements that are on the way. 

Thanks

PR #218, opened 2019-05-30
small bug fix with switch statement for selecting code snippets.

bharney-msft:


jamarw:
@bharney-msft Great catch!

PR #219, opened 2019-06-02
fix: Sync langue list with Contributing repo

nschonni:
This mainly adds new languages not in the list currently. It does drop `nodejs` that was removed (since it should be `javascript`), and it looks like the recommend `md` over `markdown`


bharney-msft:
@nschonni looks good! Thank you for the contribution. Its very much appreciated!

PR #226, opened 2019-06-05
Update preview telemetry

jamarw:
1. Update aiKey, version and changelog
2. New function to return repo information.
3. Update telemetry calls
4. New function to centralize reporter call.


docs-preview/src/extension.ts
```diff
@@ -62,3 +66,11 @@ export function codeSnippets(md, options) {
     };
     md.core.ruler.before("normalize", "codesnippet", importCodeSnippet);
 }
+
+export function sendTelemetryData(commandOption: string) {
```
adunndevster: @jamarw - Everything looks good! One thought... could this function be moved under common.ts?

jamarw: @adunndevster Great idea.  Maybe I should create a single function in the other extensions and reduce a few lines of code.

PR #227, opened 2019-06-06
93236 mut scaffold

adunndevster:


docs-metadata/src/controllers/apply-controller.ts
```diff
@@ -0,0 +1,52 @@
...
+	env, 
+	window, 
+	workspace, 
+	ExtensionContext, 
+	OpenDialogOptions,
+	Uri} from 'vscode';
+
+export async function showApplyMetadataMessage(folderPath:string)
+{
+	window.showInformationMessage(`Apply MUT file ${folderPath}?`,
```
bharney-msft: we should probably spell out the entire word instead of using MUT

adunndevster: Hmmm... didn't think about it before by I agree. This message was in the spec specifically.

docs-metadata/package.json
```diff
@@ -0,0 +1,54 @@
...
+	"activationEvents": [
+		"onCommand:extension.extract",
+		"onCommand:extension.apply"
+	],
+	"main": "./out/extension.js",
+	"contributes": {
+		"commands": [
+			{
+				"command": "extension.extract",
+				"title": "Docs Metadata: Extract"
```
jamarw: @adunndevster Everything looks great!  One question: do you think the title should be tweaked to read **Docs: Metadata extract** and **Docs: Metadata apply** to remain consistent with the other commands in the command palette?  Not a big deal at all and maybe it's different by design.  

image

adunndevster: @meganbradley - What do you think?

adunndevster: I'll change it. Easy enough to change back if needed :).

docs-metadata/src/extension.ts
```diff
@@ -0,0 +1,118 @@
...
+                const downloader = new ExtensionDownloader(logger, extension.packageJSON);
+                return downloader.installRuntimeDependencies();
+            } else {
+                return true;
+            }
+        });
+}
+
+function platformIsSupported(logger: Logger): boolean {
+	var getos = require('getos')
```
jamarw: Hi @adunndevster , there's a native node module may work here to reduce some code and drop a dependency. https://nodejs.org/dist/latest-v10.x/docs/api/os.html#os_os_platform

docs-metadata/.vscode/settings.json
```diff
@@ -0,0 +1,14 @@
...
+    "files.exclude": {
+        "out": false // set this to true to hide the "out" folder with the compiled JS files
+    },
+    "search.exclude": {
+        "out": true // set this to false to include "out" folder in search results
+    },
+    // Turn off tsc task auto detection since we have the necessary tasks as npm scripts
+    "typescript.tsc.autoDetect": "off",
+    "cSpell.words": [
+        "Synchrounous"
```
nschonni: Doesn't look like this is needed now

docs-metadata/.gitignore
```diff
@@ -0,0 +1,2 @@
+node_modules/*
+out/*
```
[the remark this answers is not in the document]
adunndevster: @jamarw thank you for fixing!

PR #269, opened 2019-07-11
103479 md025

adunndevster:


nschonni:
Is this for the issues caused by the current Tab system? That might be better handled with inline disabled rules for those documents.
Should you be overriding any repo config in the plug-in, because this won't affect any commanline execution on the documents. This can be confusing when a cli picks up something and the IDE is hiding it.

docs-markdown/src/controllers/lint-config-controller.ts
```diff
@@ -0,0 +1,32 @@
...
+
+// store users markdownlint settings on activation
+let currentMarkdownlintSettings: string[] = [];
+
+export function disableHeaderRule() {
+    // the markdownlint property we need to update
+    const markdownlintProperty = "markdownlint.config";
+    // heading value we need to disable, should append this to the current settings if they exists
+    // may want to use another array for reset
+    const headingFalse = { MD025: false };
```
bharney0: @adunndevster I think we want to keep the MD025 rule active to force users to have a title or H1 tag. But the goal is to allow both title yaml front matter and # H1. Based on this [issue](https://github.com/DavidAnson/markdownlint/issues/199) is looks like this can be done as seen [here](https://github.com/DavidAnson/markdownlint/blob/6f7c0aac131f21dd41a2a26810779c993521b733/test/hugo-quickstart-example-clean.json): 
```
"MD025": {
    "front_matter_title": ""
  }
```
instead of setting MD025 to false.

bharney0:
@nschonni this change is supposed to resolve the issue with the recent change to markdownlint extension where the MD025 rule does not allow multiple titles/headers. However, docs wants to allow both title yaml front matter and # H1 tag to be present. But the rule is now throwing a linting error that does not allow this as seen in the screen shot below.
image

nschonni:
Right, I forgot the new behaviour broke that (was thinking of the other reason this gets triggered commonly on Doc), but it's still probably better to give guidance to the repos instead of silently changing the config in VSCode only. If OpenBuild ends up integrating, you'll end up with CI failures that people can't see in VS Code. This will also affect editors on non-Docs projects if they have the extension installed

jamarw:
Opening a new PR for this.

PR #273, opened 2019-07-12
Support for MD025

jamarw:
Adding [front matter support](https://github.com/DavidAnson/markdownlint/blob/6f7c0aac131f21dd41a2a26810779c993521b733/test/hugo-quickstart-example-clean.json) to work with Markdownlint MD025 rule update.


nschonni:
Can you detect that you're running on a MS Docs project by sniffing for the openbuild or something?
This still has the problem of affecting users of the plugin whether or not their working on MS Docs projects

PR #276, opened 2019-07-16
Fix multiple top level headings lint error

jamarw:
https://github.com/DavidAnson/markdownlint/blob/master/doc/Rules.md#md025---multiple-top-level-headings-in-the-same-document

### DIFF
--- a/docs-markdown/package-lock.json
+++ b/docs-markdown/package-lock.json
@@ -4062,9 +4062,9 @@
       }
     },
     "lodash": {
-      "version": "4.17.11",
-      "resolved": "https://registry.npmjs.org/lodash/-/lodash-4.17.11.tgz",
-      "integrity": "sha512-cQKh8igo5QUhZ7lg38DYWAxMvjSAKG0A8wGSVimP07SIUEK2UO+arSRKbRZWtelMtN5V0Hkwh5ryOto/SshYIg==",
+      "version": "4.17.14",
+      "resolved": "https://registry.npmjs.org/lodash/-/lodash-4.17.14.tgz",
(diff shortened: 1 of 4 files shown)


jamarw:
Closing existing [PR](https://github.com/microsoft/vscode-docs-authoring/pull/273) that was based off develop branch.

nschonni:
Can you detect that you're running on a MS Docs project by sniffing for the openbuild or something?
This still has the problem of affecting users of the plugin whether or not their working on MS Docs projects

nschonni:
PS: I also dislike how they changed this setting, i just don't think you should override the config for all projects

jamarw:
> Can you detect that you're running on a MS Docs project by sniffing for the openbuild or something?
> This still has the problem of affecting users of the plugin whether or not their working on MS Docs projects

Hi @nschonni , good point.  Unfortunately, the setting will be added once the user opens a docs repo so the sniffing wouldn't completely solve the problem.  We'd like to remove the setting on close if the user did not set it explicitly but the deactivation process runs quickly and doesn't allow enough time to launch the reset.  A setting has been added that will allow users to disable the automated parameter setting if they'd like.  Thanks for the feedback.

nschonni:
Thanks!
One other though, if the repo has a .markdownlint.json file, they can update the config to match this. If they don't have a .markdownlint.json in the repo, there are a lot more issues that markdownlint will flag besides MD025 in a normal repo.
MD025 will also still be flagged for repos that use the "tabs" (even if that is discouraged) because they tend to use H1s in their format as well.

nschonni:
I've slowly been PRing `.markdownlint.json` files, but it's hit and miss as to which repos accept them 😉

PR #284, opened 2019-07-30
Added display Properties to xref output/preview

bharney0:


docs-preview/src/xref/xref.ts
```diff
@@ -102,7 +102,18 @@ async function updateXrefContent(md: any, src: string) {
           if (response) {
             if (response.data[0]) {
               let xref = response.data[0];
-              mdSrc = `[${xref.fullName}](${xref.href})`;
+              let displayProperty = captureGroup[3];
+              if (displayProperty) {
+                if (displayProperty == "fullName") {
+                  mdSrc = `[${xref.fullName}](${xref.href})`;
+                }
+                if (displayProperty == "nameWithType") {
```
nschonni: else if, since it can't be both?

PR #286, opened 2019-08-01
Added message to user if cannot connect to xref service, or if no sea…

bharney0:
Added message to user if cannot connect to xref service, or if no search results returned.

### DIFF
--- a/docs-markdown/src/controllers/xref-controller.ts
+++ b/docs-markdown/src/controllers/xref-controller.ts
@@ -1,32 +1,32 @@
 "use strict";
 
-import { QuickPickItem, window, CompletionItem, Range, Position } from "vscode";
+import { CompletionItem, Position, QuickPickItem, Range, window } from "vscode";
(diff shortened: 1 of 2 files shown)


docs-markdown/src/controllers/xref-controller.ts
```diff
@@ -164,10 +165,18 @@ async function getXrefDisplayProperty() {
 
 async function getXrefSelection() {
   const items: QuickPickItem[] = [];
-  let uid: string | undefined = await window.showInputBox({ placeHolder: "Enter XREF Search Term" });
+  const uid: string | undefined = await window.showInputBox({ placeHolder: "Enter XREF Search Term" });
   if (uid) {
-    let content = await getAsync(`${rootUrl}/autocomplete?tags=${tags}&text=${uid}`);
-    content.data.map((item: { tags: any; uid: string; }) => {
+    const response = await getAsync(`${rootUrl}/autocomplete?tags=${tags}&text=${uid}`);
+    if (response.status !== 200) {
```
bharney0: Main change is to add Error message response if status code is not 200, and if there are no results returned on line 175

PR #289, opened 2019-08-02
Added docs styles to docs-preview.

bharney0:


docs-preview/src/extension.ts
```diff
@@ -35,12 +35,217 @@ export function activate(context: ExtensionContext) {
...
             return md.use(require("markdown-it-include"), { root: workingPath, includeRe: INCLUDE_RE })
                 .use(codeSnippets, { root: workingPath })
-                .use(xref);
-        }
+                .use(xref)
+                .use(custom_codeblock)
+                .use(column_end)
+                .use(container_plugin, "row", {
+                    marker: ":",
+                    validate(params) {
```
adunndevster: this is really cool! Maybe we can clean this up a little? extension.ts is starting to look a little heavy...

docs-preview/src/extension.ts
```diff
@@ -35,12 +35,217 @@ export function activate(context: ExtensionContext) {
...
+
+        // Since start is found, we can report success here in validation mode
+        //
+        if (silent) { return true; }
+
+        // Search for the end of the block
+        //
+        nextLine = startLine;
+
+        for (; ;) {
```
adunndevster: Wow... I've never seen this way of using a for loop... is this similar to:
`while(nextLine < endline){...`

?

docs-preview/media/styles.css
```diff
@@ -0,0 +1,104 @@
+body, html {
+  margin-left:auto;
+  margin-right: auto;
+  max-width: 967px;
+  padding: 0;
+  font-size: 16px !important;
```
bharney0: When comparing the sizes of font in the developer console, they seem to be the same, but visually on the screen, I seems like the font should be smaller. 16px font-size does not match the published output compared to the markdown preview visually.

PR #311, opened 2019-08-30
Update README.md

meganbradley:
Change back to .txt


meganbradley:
@adunndevster  small change

PR #313, opened 2019-09-03
Prep for docs-metadata publish.

adunndevster:


docs-metadata/package.json
```diff
@@ -2,7 +2,7 @@
 	"name": "docs-metadata",
 	"displayName": "docs-metadata",
 	"description": "Simplifies editing metadata within docs.",
-	"version": "0.0.125",
+	"version": "1.0.0",
 	"publisher": "docsmsft",
 	"icon": "images/docs-logo-ms.png",
 	"preview": true,
```
jamarw: @adunndevster , should this still be preview?

adunndevster: Oh it should not be!

PR #326, opened 2019-09-12
Image Extension

bharney0:
This is new markdig functionality that is being built for the Docs Platform to enhance images using the Markdig Extensions, which is being supported by the Docs Authoring Pack. The new image syntax
complex:
```
:::image type="complex" alt-text="Lorem Ipsum" source="./path":::

:::image-end
```
content:
```
:::image type="content" alt-text="Lorem Ipsum" source="./path":::
```
icon:
```
:::image type="icon" source="./path":::
```
- [x] Image type="content"
- [x] Image type="icon"
- [x] Image type="complex"
- [x] Add loc-scope to image
- [x] Preview support
- [x] Linting support
- [x] Auto-Completion


docs-markdown/markdownlint-custom-rules/image.js
```diff
@@ -0,0 +1,84 @@
+// @ts-check
+
```
adunndevster: @bharney-msft  - Is this file just a stub? It is called image.js but it has xref rules?

adunndevster: I'm just asking because I own the task to write image linting rules :)

bharney0: Its just a stub that I was using for the task to mock out some sudo code.

adunndevster: Got it.

PR #330, opened 2019-09-19
Adding image snippets

jamarw:


docs-markdown/snippets/snippets.json
```diff
@@ -1,12 +1,37 @@
...
-    "body": ["    :::column:::\n\n    :::column-end:::"],
+    "body": [
+      "    :::column:::\n\n    :::column-end:::"
+    ],
     "description": "column"
+  },
+  "content": {
+    "prefix": "image",
+    "body": [
+      ":::image source=\"\" alt-text=\"\":::"
```
bharney0: even though type="content" is supposed to be the default. The actual json schema requires the type property. I think we should add it into the snippet with type="content"

jamarw: Sounds good @bharney0 , updated.

PR #335, opened 2019-10-01
Version bump.

adunndevster:


jamarw:
Hi @adunndevster , does the changelog need to be updated as well or is that all good as-is?

adunndevster:
@jamarw  - I updated the changelog... how does it look?

jamarw:
> @jamarw - I updated the changelog... how does it look?

Cool @adunndevster

PR #338, opened 2019-10-03
Support tilde in file path preview

jamarw:
Add tilde regex to includes.ts and codesnippet.ts 

- used to replace **~** with the repo root path so the include/snippet source file can be read
- updated path is not written back to the markdown file and surrounding text is not affected


nschonni:
I'm not sure this is going to work for a bunch of projects. The tilda is the root of the docfx rather than the project https://dotnet.github.io/docfx/tutorial/links_and_cross_references.html

jamarw:
> I'm not sure this is going to work for a bunch of projects. The tilda is the root of the docfx rather than the project https://dotnet.github.io/docfx/tutorial/links_and_cross_references.html

Hi @nschonni , thanks for the info.  This fix will address a few scenarios we have in some of our larger docs repos but we're still gathering usage data to cover additional scenarios.

Issue #282, opened 2019-07-27
Incorrect error and disabling the backspace key


guardrex:
Occasionally when I open VSC (v1.36.1 currently), the docs-markdown extension (v0.2.12) throws an error when I use the backspace key, and the backspace key doesn't remove characters from the markdown file. Each time I press the key, the error message is ...

> The docs-markdown extension only works on Markdown files.

Sometimes the error clears itself in a minute or two. I can also clear it by closing and re-opening VSC.

bharney0:
Thank you for reporting the issue. This problem has been fixed in by [removing the delete list helper](https://github.com/microsoft/vscode-docs-authoring/pull/339/files#diff-0bd763de7d6001c4ab04957287ce986aL364) functionality in this PR #339. Closing this issue out.

PR #339, opened 2019-10-07
Fixed bug with image linting and added additional options to loc-scop…

bharney0:
Fixed bug with image linting and added additional options to loc-scope image extension.


bharney0:
Referencing Open Issue #282

PR #356, opened 2019-11-13
Updated alt text validation to handle 250 characters.

bharney0:


docs-markdown/src/controllers/media-controller.ts
```diff
@@ -261,15 +261,15 @@ export function Insert(isArt: any) {
...
             vscode.window.showInputBox({
-                placeHolder: "Add alt text (up to 70 characters)",
+                placeHolder: "Add alt text (up to 250 characters)",
             }).then((val) => {
                 if (!val) {
                     getFilesShowQuickPick(isArt, "");
                     vscode.window.showInformationMessage("No alt entered or selected.  File name will be used.");
                 } else if (val.length < 70) {
                     getFilesShowQuickPick(isArt, val);
                 } else if (val.length > 70) {
```
jamarw: @bharney0 , looks good.  I think we might just need to bump the length evaluations too and everything will be good to go.

bharney0: @jamarw good call. Thanks for catching that. The code has been updated to us 250 instead of 70.

PR #384, opened 2020-01-03
Added new command/snippet. External link (opens in new tab).

IEvangelist:
Hi @meganbradley,

Here is a much smaller pull request than my last. This one simply adds a new snippet and command for creating an external link which will open in a new tab. It's based on another PR that I have over in the guidance - [here](https://github.com/MicrosoftDocs/docs-help-pr/pull/2737). It is my belief that we should strive to keep our users on our site, to avoid context switching -- external links are perfect for opening in a new tab.

This is an example of what it renders like on our platform.

image

Please note that there are several key considerations with this pull request. It includes inline HTML, with an `a` and `span` element. These elements are [allow-listed here](https://review.docs.microsoft.com/en-us/help/onboard/admin/html-whitelist?branch=master) and I detailed this as part of this other related pull request [here](https://github.com/MicrosoftDocs/azure-docs-pr/pull/97243#issuecomment-567160782). As for the security concern, please note that I'm also using the `rel="noopener"` as [recommended here](https://developers.google.com/web/tools/lighthouse/audits/noopener#recommendations).

Screenshot of it in action:

image

Here is the resulting content:

image

### DIFF
--- a/docs-markdown/package.json
+++ b/docs-markdown/package.json
@@ -94,11 +94,6 @@
         "category": "Docs"
       },
       {
-        "command": "insertExternalURL",
-        "title": "Link to external website (opens in new tab)",
-        "category": "Docs"
-      },
-      {
         "command": "applyImage",
         "title": "Image (Docs Markdown)",
         "category": "Docs"
--- a/docs-markdown/package.json
+++ b/docs-markdown/package.json
@@ -109,6 +109,11 @@
         "category": "Docs"
       },
       {
+        "command": "insertImageWithGrayBorder",
+        "title": "Image with gray border",
+        "category": "Docs"
+      },
+      {
         "command": "applyIcon",
         "title": "Icon image",
         "category": "Docs"
--- a/docs-markdown/package.json
+++ b/docs-markdown/package.json
@@ -94,6 +94,11 @@
         "category": "Docs"
       },
       {
+        "command": "insertExternalURL",
+        "title": "Link to external website (opens in new tab)",
+        "category": "Docs"
+      },
+      {
         "command": "applyImage",
         "title": "Image (Docs Markdown)",
         "category": "Docs"
--- a/docs-markdown/snippets/snippets.json
+++ b/docs-markdown/snippets/snippets.json
@@ -33,17 +33,5 @@
       ":::image type=\"complex\" source=\"\" alt-text=\"\":::\n\n:::image-end:::"
     ],
     "description": "Complex image"
-  },
-  "externalLink": {
-    "prefix": [
-      "link",
-      "ext"
-    ],
-    "body": [
-      "<a href=\"${1:url}\" target=\"_blank\" rel=\"noopener\">",
-      "    ${2:URL description} <span class=\"docon docon-navigate-external x-hidden-focus\"></span>",
-      "</a>"
-    ],
-    "description": "External link, opens in a new tab."
   }
 }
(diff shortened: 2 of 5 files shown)


meganbradley:
Thanks @IEvangelist - I'll have to discuss this one with the content leadership and engineering team. Although these elements are currently whitelisted, we have plans to further restrict HTML and instead extend Markdown for similar functionality. I don't know if there are any plans that affect this particular HTML, but before we explicitly support it in the Docs Authoring Pack I'll need to verify. CC @lamebrain @jamarw @adunndevster @adkinn

IEvangelist:
> Thanks @IEvangelist - I'll have to discuss this one with the content leadership and engineering team. Although these elements are currently whitelisted, we have plans to further restrict HTML and instead extend Markdown for similar functionality. I don't know if there are any plans that affect this particular HTML, but before we explicitly support it in the Docs Authoring Pack I'll need to verify. CC @lamebrain @jamarw @adunndevster @adkinn

Awesome, sounds good -- thank you. In the meantime, I could write the extended functionality in our Markdown engine, just point me in the direction. I'm assuming it's docfx , if so I'm already familiar with this project and would love to write both parts -- the extension here that would insert content that docfx would then turn into HTML.

meganbradley:
@IEvangelist for the image border, do you mean adding [!div class="mx-imgBorder"] ? Can you describe the experience change in more detail?

IEvangelist:
> @IEvangelist for the image border, do you mean adding [!div class="mx-imgBorder"] ? Can you describe the experience change in more detail?

Hi @meganbradley,

Yes, this addition provides the image border syntax, something that I always forget and have to look up in the [Contributor's Guide](https://review.docs.microsoft.com/en-us/help/contribute/contribute-how-to-create-screenshot?branch=master#gray-border-example). This adds it as a command palette item.

From within a markdown file, open command palette and select "Docs: Image with gray border" option:

image

Then enter required alt text:

image


Then select the available image file:

image

Then enjoy the syntax already completed for you correctly:

image

meganbradley:
@IEvangelist I think adding support for mx-imgBorder to the command palette is fine, as that syntax is supported until it's replaced with a new extension. 

I just talked to @DuncanmaMSFT about defaulting to opening external links in new tabs, and he confirmed that this is NOT recommended for the following reasons:

- Users can already open in a new tab if they want to.
- Forcing them to open in a new tab means they lose context, and can't easily retrace their steps. This is particularly a problem for users of screen readers and users reading on mobile devices.

Can you break out the image border change so it can be reviewed separately?

CC @lamebrain @adkinn @tysonn

IEvangelist:
Hi @meganbradley,

Done, I've updated this PR to remove the external link functionality but kept the image border functionality. Thank you

adkinn:
@meganbradley @IEvangelist this is definitely old syntax I wish we would have replaced by now. It shouldn't be that much work to add a border property to the new image extension. I'd rather we promote the new Image extension then use this old method.

meganbradley:
@adkinn do you think we can get image updates in this quarter?

adkinn:
@meganbradley I think it's a very small bit of work but let's sync with Brian and see.

IEvangelist:
> @meganbradley @IEvangelist this is definitely old syntax I wish we would have replaced by now. It shouldn't be that much work to add a border property to the new image extension. I'd rather we promote the new Image extension then use this old method.

How would an outside contributor on this have known this was old syntax?

PR #391, opened 2020-01-13
Fixed a bug where directories under include folders were not found

IEvangelist:
Hi @meganbradley,

I was experiencing a bug where markdown files were not being displayed in the quick selection list when using the <kbd>F1</kbd> Docs: Include functionality. The issue is that the glob pattern didn't account for files potentially nested under an includes directory. For example, prior to this change the only markdown files displayed in the quick selection would be the `include-file.md` whereas, I was expected all the markdown files under the includes directory.

```
└── includes
     ├── subfolder
     │    ├── sub-file-01.md
     │    └── sub-file-02.md
     └── include-file.md
```

With this change, all of these markdown files are displayed.

### DIFF
--- a/docs-markdown/package-lock.json
+++ b/docs-markdown/package-lock.json
@@ -496,6 +496,23 @@
         "@babel/types": "^7.3.0"
       }
     },
+    "@types/events": {
+      "version": "3.0.0",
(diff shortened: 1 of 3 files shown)


jamarw:
Hi @IEvangelist , thanks for the update.  Checking it out now.

jamarw:
@IEvangelist , looks good!

PR #397, opened 2020-01-22
Added the ability to insert links from yaml files.

IEvangelist:
Hi @meganbradley,

I have added the ability to insert links from yaml files. This is HUGELY helpful with links that we create in `index.yml` and `toc.yml` files. Today we do not have support in place for this functionality. Here is an example of it working...

yaml-links

### DIFF
--- a/docs-markdown/package.json
+++ b/docs-markdown/package.json
@@ -4,7 +4,7 @@
   "description": "Docs Markdown Extension",
   "icon": "images/docs-logo-ms.png",
   "aiKey": "0a0e5961-85c2-451a-bce8-6a54e37c93be",
-  "version": "0.2.33",
+  "version": "0.2.34",
   "publisher": "docsmsft",
   "homepage": "https://github.com/Microsoft/vscode-docs-authoring/tree/master/docs-markdown",
   "bugs": {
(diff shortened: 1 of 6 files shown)


adunndevster:
I tested this locally and it works great.

@meganbradley - What are your thoughts?

meganbradley:
@adunndevster based on the gif it looks good. I see no problem with merging it as is, but I'd like to review the experience and probably update the YAML menu so it's consistent with inserting links in TOC files.

IEvangelist:
> @adunndevster based on the gif it looks good. I see no problem with merging it as is, but I'd like to review the experience and probably update the YAML menu so it's consistent with inserting links in TOC files.

Are you referring to the <kbd>Alt+M</kbd> menu? If so, @adunndevster and I discussed that a bit. I could probably add it quick.

meganbradley:
@ievangelist @adunndevster yes - I'd like to see it in Alt+M for YAML files only. Thanks!

IEvangelist:
> @IEvangelist @adunndevster yes - I'd like to see it in Alt+M for YAML files only. Thanks!

Done, the latest addition adds it to the <kbd>Alt+M</kbd> menu for "Insert link"

PR #398, opened 2020-01-23
Introduced markdown completion and validation for code-fence language ids

IEvangelist:
Hi @meganbradley,

Here is another pull request, but this one adds markdown completion and validation for code-fence language identifiers. This is imperative to help ensure that content developers are using the correct and available language identifiers. If they are not, it will not render the correct syntax highlighting when published.

## Auto-completion

auto-complete

## Validation & correction

validate-langs

## New settings option

This is to allow the user to configure whether or not the suggestions include all the available languages, be default it is off. By default we're only showing popular languages, that are common in our docs -- but again, the user can opt-in to showing all languages.

md-config


adunndevster:
@IEvangelist  - Another amazing feature! Quick thought, when I put my closing tick marks in, I still get the autocompletion prompt. Is there any way to detect if they are closing ticks vs opening ones?

IEvangelist:
> @IEvangelist - Another amazing feature! Quick thought, when I put my closing tick marks in, I still get the autocompletion prompt. Is there any way to detect if they are closing ticks vs opening ones?

I could try to detect that but I'm not sure how that would work. It starts to get complex, as you might imagine. When you first start typing a code fence, you'd want the completion options which would be in the same state as the closing one... do you have any ideas?

adunndevster:
Hmmm... lemme think. Is there any other time in markdown when triple-ticks are used besides code fences?

If not, then maybe...
- look at the whole doc and determine if this set of ticks is an odd set rather than even
- if it is odd, run the autocompletion

Another thought could be...
- before showing the autocompletion, search backwards and see if there is a previous set of ticks
- if there are, does it match something like this? /```\w+/
- if so, don't show the autocompletion
- if no ticks are found, or ticks are found that don't match that value, render the autocompletion

Just brainstorming. You know a lot more about this current puzzle than I do, so let me know what you think :)

IEvangelist:
I figured it out, I'm made assumptions about the ordinal position of code fences. It works though! :shipit:

adunndevster:
@meganbradley - I tried this feature out locally and it seems to work great. Any thoughts on it before we merge and publish the feature?

meganbradley:
@adunndevster nope - go for it! I haven't had a chance to try it out but I can always request tweaks later. Don't wait on me. :)

adunndevster:
Thanks @meganbradley !

PR #399, opened 2020-01-24
Added the ability to sort selections: asc or desc.

IEvangelist:
Hi @meganbradley,

I added the ability to sort selections, either ascending or descending. This is helpful to ensure that certain things are sorted correctly and useful for tables, in fact the reason I created this is due to the manual difficulty I was having when trying to sort language and locale tables for Speech service docs. I really could have used this feature then :smile:.

## Sort selection
sort-selected-lines

> **Note:** the menu item is only ever available when the editor has a selection.

When testing this locally, feel free to use this markdown -- or any other variation.

```markdown
| Letter | Details |
|--------|---------|
| Z | The last letter in the alphabet |
| C | The a letter after A in the alphabet |
| X | The alphabet letter is towards the end |
| M | Somewhere in the middle? |
| A | The first letter in the alphabet |
```

### DIFF
--- a/docs-markdown/package.json
+++ b/docs-markdown/package.json
@@ -4,7 +4,7 @@
   "description": "Docs Markdown Extension",
   "icon": "images/docs-logo-ms.png",
   "aiKey": "0a0e5961-85c2-451a-bce8-6a54e37c93be",
-  "version": "0.2.34",
+  "version": "0.2.36",
   "publisher": "docsmsft",
   "homepage": "https://github.com/Microsoft/vscode-docs-authoring/tree/master/docs-markdown",
   "bugs": {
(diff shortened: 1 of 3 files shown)


IEvangelist:
> Hey @IEvangelist , this looks really good. One case that I wanted your opinion on since you are aware of more authoring scenarios than I am is numbering. If you sort the markdown below, 11 will show up before 2. Do you think this is a common scenario or just an edge case? Thanks!
> 
> | 1 | Number 1 |
> | Aa | The first letter in the alphabet |
> | Ab | The first letter in the alphabet |
> | C | The a letter after A in the alphabet |
> | M | Somewhere in the middle? |
> | 2 | Number 2 |
> | X | The alphabet letter is towards the end |
> | Z | The last letter in the alphabet |
> | 11 | Number 11 |

That's a really good catch! I'll fix that, I had forgotten about natural language sorting.

PR #406, opened 2020-01-29
Remove empty metadata for cleanup scripts

jamarw:
To-do's
1. Add metadata attributes
2. Delete empty lines


docs-markdown/src/helper/cleanup.ts
```diff
@@ -0,0 +1,101 @@
+"use strict";
+
+export function deleteEmptyMetadata(data: any) {
+    const msAuthorRegex: any = new RegExp(/(ms\.author:)\s([a-zA-Z0-9]|\[)/i);
+    const authorRegex: any = new RegExp(/(author:)\s([a-zA-Z0-9]|\[)/i);
```
IEvangelist: The authorRegex will capture both the `author` and the `ms.author`. I had the same issue, I opted to use the beginning of the line indicator `^`, [see here](https://github.com/microsoft/vscode-docs-authoring/pull/407/files#diff-0fbd6f34b1f944104a1f9e67209df180R69).

jamarw: Hey @IEvangelist , good catch!  Thought I changed all the regex to start at the beginning but missed this section.

PR #407, opened 2020-01-30
Added implicit metadata updates

IEvangelist:
Hi @meganbradley,

I've got another nice-to-have feature that will hopefully help to increase content development productivity. I just wanted to add a note here explaining how appreciative I am of this amazing platform, for being so accepting of contributions and supportive. I feels great to collaborate on these kinds of things!

See details below:

Watch this short GIF, that demonstrates this new functionality. Notice the various metadata values, such as the `titleSuffix`, `author`, `manager`, `ms.service`, `ms.subservice`, `ms.date` and `ms.author`. All of these values are discoverable in the workspace `docfx.json` if specified. If they are not found there, the `author` and `ms.author` are pulled from the templates config. What this means, is that for repos that have configured implicit metadata values -- authors can now trigger implicit metadata updates if the metadata values are explicitly set to something different.

## Update Implicit Metadata Values

implicit-metadata-updates

### Notes

This design is based on the [`docfx.json` file format](https://dotnet.github.io/docfx/tutorial/docfx.exe_user_manual.html#3-docfxjson-format) and executes on a discovery-based approach. When there is a `docfx.json` file present, we parse it, and evaluate various metadata values. These values are then used as potential replacements, the reason they're only potential replacements is that we only replace metadata values that are present in the contextual markdown file.

### DIFF
--- a/docs-markdown/package.json
+++ b/docs-markdown/package.json
@@ -201,8 +201,8 @@
         "category": "Docs"
       },
       {
-        "command": "updateAllMetadataValues",
-        "title": "Update All Metadata Values",
+        "command": "updateImplicitMetadataValues",
+        "title": "Update Implicit Metadata Values",
         "category": "Docs"
       }
     ],
--- a/docs-markdown/package.json
+++ b/docs-markdown/package.json
@@ -179,8 +179,8 @@
         "category": "Docs"
       },
       {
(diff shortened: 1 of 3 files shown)


IEvangelist:
#sign-off

PR #408, opened 2020-01-30
Added red tiger

adunndevster:


docs-markdown/src/controllers/media-controller.ts
```diff
@@ -47,14 +47,15 @@ export function insertVideo() {
         return;
     } else {
         vscode.window.showInputBox({
-            placeHolder: "Enter URL; https://channel9.msdn.com or https://www.youtube.com is a required prefix for video URLs",
-            validateInput: (urlInput) => urlInput.startsWith("https://channel9.msdn.com") && urlInput.split("?")[0].endsWith("player") ||
-                urlInput.startsWith("https://www.youtube.com/embed") ? "" :
-                "https://channel9.msdn.com or https://www.youtube.com/embed are required prefixes for video URLs. Link will not be added if prefix is not present.",
+            placeHolder: "Enter URL; Begin typing to see the allowed video URL prefixes.",
+            validateInput: (urlInput) => urlInput.toLowerCase().startsWith("https://channel9.msdn.com") && urlInput.toLowerCase().split("?")[0].endsWith("player") ||
```
IEvangelist: Although this is a minor suggestion (nit), it would be more efficient to only call the `urlInput.toLowerCase()` once.

```typescript
const validateInput = (urlInput: string) => {
    const urlLowerCase = urlInput.toLowerCase();
    return urlLowerCase.startsWith("https://channel9.msdn.com")
        && urlLowerCase.split("?")[0].endsWith("player")
        || urlLowerCase.startsWith("https://www.youtube.com/embed")
        || urlLowerCase.startsWith("https://www.microsoft.com/en-us/videoplayer/embed")
        ? ""
        : "https://channel9.msdn.com, https://www.youtube.com/embed or https://www.microsoft.com/en-us/videoplayer/embed are required prefixes for video URLs. Link will not be added if prefix is not present."
};
const val = await vscode.window.showInputBox({
    placeHolder: "Enter URL; Begin typing to see the allowed video URL prefixes.",
    validateInput
});
if (val === undefined) {
    postWarning("Incorrect link syntax. For YouTube videos, use the embed syntax, https://www.youtube.com/embed/<videoID>. For Channel9videos, use the player syntax, https://channel9.msdn.com/. For Red Tiger, use, https://www.microsoft.com/en-us/embed/<videoID>/player");
    return;
}
const contentToInsert = videoLinkBuilder(val);
```

adunndevster: @IEvangelist - Good call :). I got a bit lazy there... tidying up!

PR #412, opened 2020-02-10
Add .NET Core CLI dev lang to highlight languages list

scottaddie:
The `dotnetcli` dev lang was missing. Adding it in this PR and setting as popular so that it appears in the triple-tick autocomplete dialog. This dev lang is used extensively across multiple doc sets. Example: https://docs.microsoft.com/aspnet/core/client-side/libman/libman-cli?view=aspnetcore-3.1#installation


scottaddie:
Note that I've also created the following PR to update Duncan's README file: https://github.com/DuncanmaMSFT/highlight.js/pull/27

PR #414, opened 2020-02-12
Fixed a bug where image compression was incorrectly assuming plugin.

IEvangelist:


docs-images/src/compressor.ts
```diff
@@ -127,20 +130,49 @@ export class ImageCompressor {
     }
 
     private async tryApplyImageCompression(filePath: string) {
+        const plugins = [];
+        const fileExtension = this.getFileExtension(filePath);
+        switch ((fileExtension || "").toLowerCase()) {
+            case ".png":
+                plugins.push(
+                    imageminPng({   // https://www.npmjs.com/package/imagemin-optipng#optimizationlevel
```
jamarw: @IEvangelist , super small casing comment.  Should this be image**In**Png?  This would be the same for the other cases if you think the casing should be updated.  There's no impact on functionality.

IEvangelist: This is actually the imported `imageminPng` plugin. It stands for "image min png". It is correct as is and wouldn't work if the casing was changed.

PR #417, opened 2020-02-13
Enhance code fence auto-completion in docs-markdown extension

scottaddie:
Summary of changes:

- Fix the truncation of the code block closing (```) in the auto-complete item text. There was a `// This is missing from the completion item!` comment in the code to track this issue.
- After selecting a dev lang from the auto-complete list, do the following:
	- Close the code block (via ```)
	- Position the cursor inside of the code block

Preview with the changes in this PR:

devlangs


docs-markdown/src/helper/highlight-langs.ts
```diff
@@ -301,17 +302,17 @@ function getLanguageIdentifierCompletionItems(range: Range | undefined, isCancel
...
-                const langId = item.insertText = lang.aliases[0];
-                const doc = new MarkdownString(`Use the "${lang.language}" language identifer (alias: ${item.insertText}).`);
-                doc.appendText("\nSample:\n");
-                doc.appendCodeblock(`\`\`\`${langId}`, "markdown");
-                doc.appendCodeblock("// some code...", "markdown");
-                doc.appendCodeblock("\n", "markdown");
-                doc.appendCodeblock("```", "markdown"); // This is missing from the completion item!
-                doc.appendText("");
-                item.documentation = doc;
+
```
IEvangelist: What happens if there is already a closing ` ``` ` code fence?

scottaddie: Great question. The code doesn't attempt to handle that scenario. Should it?

IEvangelist: I think it's simply something to consider, I'm not certain - but I do know that other areas in this tooling set do take into account things like this to make the extensions a bit more adaptive. An example is here, where we can take into account whether the user has made a selection or not and base how to do something on that. https://github.com/microsoft/vscode-docs-authoring/blob/master/docs-markdown/src/controllers/image-controller.ts#L126

scottaddie:
@meganbradley @jamarw Please review when you can.

scottaddie:
@bharney0  Is this ready to merge?

PR #421, opened 2020-02-21
Feature preview mode

jamarw:


docs-markdown/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Docs Markdown Extension",
   "icon": "images/docs-logo-ms.png",
   "aiKey": "0a0e5961-85c2-451a-bce8-6a54e37c93be",
-  "version": "0.2.50",
+  "version": "0.2.51",
```
lamebrain: Let's sneak this in with the next release since there isn't any tangible change.

jamarw: @lamebrain , sounds good!

PR #429, opened 2020-02-26
Added the ability to replace smart quotes in markdown files.

IEvangelist:
Hi @meganbradley,

Here is a feature that I believe will be super impactful. It will automatically replace smart quotes in MD files. There is a very common workflow where PMs provide Word docs with content and if authors copy & paste, this leads to MD files being littered with undesirable characters. Here is a quick demo of this feature in action (of course backed by a configuration making it optional).

When `true`, automatically replace smart quotes (`“, ”, ‘, and ’`) such as those found in Word documents - with standard quotes (`" and '`) in markdown files.

replace-smart-quotes-ez

### DIFF
--- a/docs-markdown/package.json
+++ b/docs-markdown/package.json
@@ -4,7 +4,7 @@
   "description": "Docs Markdown Extension",
   "icon": "images/docs-logo-ms.png",
   "aiKey": "0a0e5961-85c2-451a-bce8-6a54e37c93be",
-  "version": "0.2.53",
+  "version": "0.2.54",
   "publisher": "docsmsft",
   "homepage": "https://github.com/Microsoft/vscode-docs-authoring/tree/master/docs-markdown",
   "bugs": {
--- a/docs-markdown/package.json
+++ b/docs-markdown/package.json
@@ -463,7 +463,7 @@
         "markdown.replaceSmartQuotes": {
           "type": "boolean",
           "default": true,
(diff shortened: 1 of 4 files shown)


bharney0:
@IEvangelist I just pull the code down and it seems to work really well. I noticed one small issue with special characters as the first character in the document. It does not seem to replace the very first smart quote for example when it starts the document? Can you see what we can do about this issue? Other wise everything looks good.
image

IEvangelist:
> @IEvangelist I just pull the code down and it seems to work really well. I noticed one small issue with special characters as the first character in the document. It does not seem to replace the very first smart quote for example when it starts the document? Can you see what we can do about this issue? Other wise everything looks good.
> image

Great catch, the issue was related to as slight misconception with how coalescing worked.

```typescript
const value: number | undefined = 0;
const index = value || -1;
```

If you thought `index` was `0` (like I did), it's actually `-1`. Here is [the correction](https://github.com/microsoft/vscode-docs-authoring/pull/429/commits/804f5284fa3a79f38c78c404387af79bcfd00f84#diff-46a6af4dda9b9aa0246a0c1d3f962244R422)...

```typescript
const index = result.index !== undefined ? result.index : -1;
```

bharney0:
@IEvangelist it looks like the npm command `npm vscode:prepublish` doesnt like the coalesce operator. Not sure why, but its throwing an error in the CI pipeline:
```
src/helper/utility.ts:422:45 - error TS1109: Expression expected.

422                 const index = result.index ?? -1;
                                                ~

src/helper/utility.ts:422:49 - error TS1005: ':' expected.

422                 const index = result.index ?? -1;
                                                    ~
```

Issue #212, opened 2019-05-07
Move custom Markdownlint rules to separate packages


nschonni:
The custom markdownlint rules are pretty helpful https://github.com/microsoft/vscode-docs-authoring/tree/develop/docs-markdown/markdownlint-custom-rules
Because they're internal to the extension, they can't be run from the CLI though. Could they be moved to a separate NPM package so they can be added to the individual repo's `.markdownlint.json`?

jamarw:
Hi @nschonni , good idea.  Passing on to PM team.

nschonni:
@DavidAnson do you have any advice for this? It might be a good example case if you can help move them, since it also interacts with VS Code plug-ins

DavidAnson:
I think it should be fairly straightforward. There’s a really basic sample here: https://github.com/DavidAnson/markdownlint/tree/master/test/rules/npm

If there are any issues, I’m happy to help!

bharney0:
This work is being done on this PR #432. Should be merged shortly as docs-linting VSCode extension.

DavidAnson:
@bharney0, that PR seems to move the custom linting rules into a new VS Code extension. I think the intent of this issue was to expose those rules as dedicated npm packages for easier reuse. Is that still under consideration?

bharney0:
@DavidAnson thats a good point. I think when our team was evaluating that, we were not sure others would use such an npm package. But maybe we were selling ourselves short. Im going to keep this issue open and hopefully we can revisit this by making a docs-linting npm package in the future.

DavidAnson:
Cool, thank you all!

nschonni:
Here is an example. Right now, the dotnet/docs run Markdownlint as a GitHub Action https://github.com/dotnet/docs/blob/master/.github/workflows/markdownlint.yml
If theses where distributed as standalone packages, they could also be enforced by CI as well as from VS Code.

PR #432, opened 2020-02-26
175851 docs linting

adunndevster:


docs-linting/src/extension.ts
```diff
@@ -0,0 +1,79 @@
...
+}
+
+/**
+ * Method to check for the docs custom markdownlint value.
+ * Checks for markdownlint.customRules property.  If markdownlint isn't installed, do nothing.  If markdownlint is installed, check for custom property values.
+ */
+export function checkMarkdownlintCustomProperty() {
+    const { msTimeValue } = generateTimestamp();
+    const customProperty = "markdownlint.customRules";
+    const customRuleset = "{docsmsft.docs-linting}/markdownlint-custom-rules/rules.js";
```
jamarw: Hi @adunndevster , do you think we should add logic to delete the custom ruleset entry for docs-markdown as well?  I don't think it's got any negative impact aside from noise in the output window.  Thanks

adunndevster: Ah! That's what was causing that. @jamarw - you are a wizard.

adunndevster: @jamarw - Okay I've updated this. The first time you run docs-linting you will see a little mess... it tries to load the old rules, is unable to, then refreshes the cache, and loads fine. After that the messages are clean up. Is this okay?

bharney0:
#212

docs-linting/src/extension.ts
jamarw: @adunndevster Looks great!

Issue #441, opened 2020-03-04
Constant Error Message When Editing Non-Markdown Files


acgritt:
I don't know if this is related to Issue #282 or not but whenever I am editing a CSharp file in VSCode (v1.42.1) and docs-markdown (v0.2.51) is enabled an error message pops up saying:

> The docs-markdown extension only works on Markdown files.

If I close the pop-up and press the backspace key it comes back up. The backspace key isn't disabled, I just keep getting the message when working on non-markdown files.

bharney0:
@acgritt Thank you for reporting the problem. There was a recent change that checks if the file is a markdown document on save. I think we can avoid the notification to reduce the noise. Work is being done on #442

PR #442, opened 2020-03-04
Added common method to check Markdown without notification to user.

bharney0:
Issue tracking on #441. Users are getting notified for editing files that are not markdown. 
image


bharney0:
Another user reported same issue #448

Issue #410, opened 2020-02-03
Linking behavior mixup


roygara:
Also noticed this after updating to the latest docs authoring pack.

The new link menu looks pretty good and it seems we got some extended functionality here, which is great! But I think some wires got crossed in the updates.

The behavior used to be: when using link to file in repo, it would simply take the article title. This makes sense because (as far as I know) standard practice is to take the article name for the link. But now it no longer does this, instead it prompts you for alt text, which you can ignore and it'll then populate with the name.

And now, when you choose to link to a web page, it doesn't prompt you for the alt text (it used to), instead it'll just automatically take the full URL and put that in for the display text as well. This doesn't make sense to me, since our standard practice was to use alt text instead of the URL for external links.

I think these two behaviors mistakenly got flipped but, if this is intentional, feel free to ignore this.

bharney0:
@roygara thank you for reporting these 2 problems. We have file a bug in Azure DevOps to track the dev work. We should get to this within the next week or so. 

1. [Link to file in Repo Bug](https://dev.azure.com/ceapex/Engineering/_workitems/edit/180568)
2. [Link to web pack Bug](https://dev.azure.com/ceapex/Engineering/_workitems/edit/183067)

jamarw:
Hi @roygara , this has been fixed in the latest release of the docs-markdown extension (v.0.2.57).  Thanks

PR #451, opened 2020-03-05
Update link to web flow

jamarw:
This PR will clean up support for three scenarios:

1. **Selected text**: If text is selected the `link to web` command will not prompt for link text and use the selected text.
2. **No text selected and link text**: If no text is selected, the user will be asked for link text. If text is provided, it will be used to construct the link.
3. **No text selected and no link text**: If no text is selected, the user will be asked for link text. If no text is entered, the url will be used to construct the link.


jamarw:
Addresses second work item in issue #410

PR #455, opened 2020-03-10
Expanded the usage of the update metadata values command to YAML

IEvangelist:
I wanted to use some of the update metadata commands now for a while in YAML, as they are the same in terms of replacing matched `RegExp` in the `metadata-controller.ts` logic. This pull request does enables it for YAML now too.


bharney0:
@IEvangelist I'll try to get this released and into production today.

PR #460, opened 2020-03-13
185524 mocha

adunndevster:


docs-markdown/__mocks__/vscode.js
```diff
@@ -0,0 +1,73 @@
+const languages = {
```
jamarw: Hey @adunndevster , do we still need the mocks directory if we're moving away from Jest?

docs-markdown/package.json
```diff
@@ -652,9 +652,10 @@
   "scripts": {
     "vscode:prepublish": "tsc -p ./",
     "compile": "tsc -watch -p ./",
-    "postinstall": "node ./node_modules/vscode/bin/install",
-    "test-core": "node ./node_modules/vscode/bin/test",
     "test": "jest",
```
jamarw: Should the test script be switched to run mocha instead of jest?

docs-markdown/src/helper/__mocks__/telemetry.js
```diff
@@ -0,0 +1 @@
+export const reporter = jest.fn();
```
jamarw: Still need this one?

docs-markdown/tsconfig.json
```diff
@@ -19,6 +19,9 @@
         "src/**/*",
         "test/**/*"
     ],
+    "setupFiles": [
+        "__mocks__/*"
```
jamarw: Still needed?

docs-markdown/__mocks__/vscode.js
adunndevster: Good point... we don't.

adunndevster: Fixed!

docs-markdown/package.json
adunndevster: Fixed!

docs-markdown/tsconfig.json
adunndevster: Fixed!

PR #477, opened 2020-03-23
Implemented daisy chain resolution

IEvangelist:
This pull request adds the ability for daisy chains to be resolved in the *.openpublishing.redirection.json* file automatically. This details of this feature were initially discussed here as [considerations](https://docs.microsoft.com/en-us/contribute/docs-authoring/sort-redirects#considerations).

redirect-resolution

## Example

Below is a before and after scenario, to demonstrate the goal of this functionality.

### Before

```json
{
    "redirections": [
        {
            "source_path": "articles/subject/a.md",
            "redirect_url": "/azure/subject/b",
            "redirect_document_id": false
        },
        {
            "source_path": "articles/subject/b.md",
            "redirect_url": "/azure/subject/c?query-string#bookmark",
            "redirect_document_id": false
        },
        {
            "source_path": "articles/subject/c.md",
            "redirect_url": "/azure/subject/d?pivot=example",
            "redirect_document_id": true
        },
        {
            "source_path": "articles/subject/z.md",
            "redirect_url": "https://docs.microsoft.com/azure/subject/b",
            "redirect_document_id": false
        }
    ]
}
```

### After

```json
{
    "redirections": [
        {
            "source_path": "articles/subject/a.md",
            "redirect_url": "/azure/subject/d"
        },
        {
            "source_path": "articles/subject/b.md",
            "redirect_url": "/azure/subject/d"
        },
        {
            "source_path": "articles/subject/c.md",
            "redirect_url": "/azure/subject/d?pivot=example",
            "redirect_document_id": true
        },
        {
            "source_path": "articles/subject/z.md",
            "redirect_url": "/azure/subject/d"
        }
    ]
}
```

If the user initiates this command with a file other than the *.openpublishing.redirection.json* file, they'll see an error like this:

image

Additionally, there are settings that need to be specified. The **Docset Name** and **Docset Root Folder Name** are required in order for the tool to understand the route segments in the URLs.

image

### DIFF
--- a/docs-markdown/package.json
+++ b/docs-markdown/package.json
@@ -458,10 +458,24 @@
       "type": "object",
       "title": "Docs Markdown Extension Configuration",
       "properties": {
+<<<<<<< HEAD
         "markdown.replaceSmartQuotes": {
           "type": "boolean",
           "default": true,
           "markdownDescription": "When `true`, automatically replace smart quotes (`“, ”, ‘, and ’`) with standard quotes (`\" and '`) in markdown files. This feature extends to other characters as well, i.e.; (`©, ™, ®, •`, subscript and superscript characters). This is useful when pasting text from Word documents."
+=======
+        "markdown.docsetName": {
+          "type": "string",
+          "default": "azure",
+          "markdownDescription": "The name of the docset, most often the route segment immediately following the locale in the URL, (i.e.; https://docs.microsoft.com/en-us/azure/ would be 'azure'). Sometimes this value is specified in the `docfx.json` file under the `build/dest` node.",
+          "scope": "window"
+        },
+        "markdown.docsetRootFolderName": {
+          "type": "string",
+          "default": "articles",
+          "markdownDescription": "The name of the docset root folder. Sometimes this value is specified in the `docfx.json` file under the `build/resource/src` node.",
+          "scope": "window"
+>>>>>>> Initial bits for handling daisy chains in redirects.
         },
         "markdown.allAvailableLanguages": {
           "type": "boolean",
--- a/docs-markdown/package.json
+++ b/docs-markdown/package.json
@@ -44,6 +44,11 @@
         "category": "Docs"
       },
       {
+        "command": "applyRedirectDaisyChainResolution",
+        "title": "Resolve redirection file daisy-chains",
+        "category": "Docs"
+      },
+      {
         "command": "formatBold",
         "title": "Bold",
         "category": "Docs"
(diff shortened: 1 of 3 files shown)


bharney0:
@IEvangelist Its strange, when I initially tried testing this, I opened up the `.openpublishing.redirection.json` file, then run `Docs: Resolve redirection file daisy-chains` I get an error for command not found. But if I open up a markdown document and run it it works fine. Not sure what the problem was. 

To repro I had to close my markdown documents, first open `.openpublishing.redirection.json` and then try to run the command. Then I opened an markdown document, and ran the command again without error. Can you see if anything can be done about that?

image

image

IEvangelist:
> @IEvangelist Its strange, when I initially tried testing this, I opened up the `.openpublishing.redirection.json` file, then run `Docs: Resolve redirection file daisy-chains` I get an error for command not found. But if I open up a markdown document and run it it works fine. Not sure what the problem was.
> 
> To repro I had to close my markdown documents, close my instance of VSCode. Then re-open VSCode and first open `.openpublishing.redirection.json` and then try to run the command. Can you see if anything can be done about that?
> 
> image
> 
> image

I think this is a limitation with extension, as its only available when the markdown extensions are activated. Perhaps, it belongs in a different extension -- at is works against the open publishing JSON file. But there was other redirect related bits so I kind of felt like it did belong here.

We could move all of them to perhaps a new extension that works against assumptions of DocFX and Open Publishing, we could name that one **docs-openpublishing**?  What do you think?

bharney0:
I think we can keep it in one extension, but if we get more features that are along the same lines maybe we can think about separating them into another extension in the future. I guess we will have to live with the limitation of opening a markdown document.

bharney0:
After some testing, everything seemed to work well! The feedback to the user that they need to navigate to the redirection file helps a lot. And the output looks good.

One question, so if the user is working between multiple repos, they will need to change the Docsset and Docset root folder, but if they forget to change it, will the program still work as expected?

IEvangelist:
> @IEvangelist After some testing, everything seemed to work well! The feedback to the user that they need to navigate to the redirection file helps a lot. And the output looks good.
> 
> One question, so if the user is working between multiple repos, they will need to change the Docsset and Docset root folder, but if they forget to change it, will the program still work as expected?

The beauty of this configuration is that the average user will never have to touch it. We source control the **.vscode/settings.json** file at the repo level. This means we will add it once and it will be picked up automatically for all users implicitly. The same is true with how manage the `markdown.docsetLanguages` settings.

PR #478, opened 2020-03-23
Fix TSLint errors in docs-markdown helpers and extension.ts

jamarw:


nschonni:
Might be worth looking at migrating to ESLint later, since TSLint is deprecated https://github.com/palantir/tslint/issues/4534

PR #521, opened 2020-04-01
Bundle docs-preview with webpack

bharney0:
Bundle docs-preview with webpack


docs-preview/.vscodeignore
```diff
@@ -1,11 +1,15 @@
...
+src/
 .gitignore
 tsconfig.json
 .markdocs/**
 install.Lock
-install.Begin
\ No newline at end of file
+install.Begin
+webpack.config.js
+images/howto.gif
```
jamarw, line 12: Hey @bharney0 , can we wildcard the images directory so we don't have to add the individual files?

bharney0: We actually need one of the image files for the images/docs-logo-ms.png. Wasnt sure how to exclude everything except the logo

jamarw: Ah, got it.

Issue #483, opened 2020-03-24
Unit Test Coverage: italic-controller.ts


bharney0:
Add unit test coverage for italic-controller

PR #560, opened 2020-04-29
Unit Test Coverage: italic-controller.ts #483

rsknutson:
Issue #483


bharney0:
@rsknutson Thank you for your contribution to Docs Authoring Pack. We'll be looking into the unit test coverage and let you know if we have feedback.

docs-markdown/src/test/suite/controllers/italic-controller.test.ts
```diff
@@ -0,0 +1,82 @@
...
+        formatItalic();
+        await sleep(300);
+        expect(spy).to.have.been.called();
+        stub.restore();
+    });
+    test("Italic Format Multiple Selection", async () => {
+        const filePath = resolve(__dirname, "../../../../../src/test/data/repo/articles/docs-markdown.md");
+        await loadDocumentAndGetItReady(filePath);
+        const editor = window.activeTextEditor;
+        common.setSelectorPosition(editor!, 159, 0, 159, 4);
```
bharney0, line 73: Can you use `editory.selections` so we are doing multiple selections. Currently this just selects one selection. But we want to test multiple selections. See here https://code.visualstudio.com/api/references/vscode-api#TextEditor `Selection[]`. You should be able to do this by doing 
```
editor.selections = [
    new Selection(fromPosition, toPosition), 
    new Selection(secondFromPosition, secondToPosition)
]
```

rsknutson: Requested usage of Selection is now implemented

bharney0: looks good, thanks for the updated changes.

PR #568, opened 2020-05-05
Added triple colon completion provider for extensions.

bharney0:
When user does `:::` we provide options for the triple colon extensions. And when user selects one of the options we provide the output snippet for that extension.

image


docs-markdown/src/helper/triple-colon-extensions.ts
```diff
@@ -0,0 +1,109 @@
...
+    {content}
+:::column-end:::`,
+        extension: "Column :::column:::",
+    },
+    {
+        content: `image type="content" source="{source}" alt-text="{alt-text}":::`,
+        extension: "Image :::image:::",
+    },
+    {
+        content: `moniker range="{range}"
```
lamebrain, line 26: When I use moniker, the indenting is off:
image

bharney0: Fixed the indent on moniker option.
image

PR #570, opened 2020-05-06
Working on coverage fix.

adunndevster:


docs-markdown/src/test/suite/index-coverage.ts
```diff
@@ -31,9 +34,7 @@ function setupNyc() {
 }
 
 export function run(): Promise<void> {
-
-	const nyc = setupNyc();
-
+	let nyc = setupNyc();
```
bharney0, line 37: Do we need this `nyc` if its also at the bottom there inside `mocha.run`?

docs-markdown/src/test/suite/index-coverage.ts
```diff
@@ -61,18 +62,18 @@ export function run(): Promise<void> {
 					if (failures > 0) {
 						e(new Error(`${failures} tests failed.`));
 					} else {
+						nyc = setupNyc();
```
adunndevster, line 65: Running this again here correctly wraps the tested code files with nyc magic, and then spits out an _accurate_ report.

docs-markdown/src/test/suite/index-coverage.ts
adunndevster, line 37: Settung up nyc here sets up the initial framework of files in the out folder. Without this, line 65 below crashes.

docs-markdown/src/test/suite/index-coverage.ts
```diff
@@ -22,6 +24,7 @@ function setupNyc() {
 		reporter: ["text", "html", "cobertura"],
 		require: [
 			"ts-node/register",
+			"source-map-support/register",
```
bharney0, line 27: Do we need to add the `source-map-support` package?

docs-markdown/src/test/suite/index-coverage.ts
adunndevster: @bharney0 - See this to answer your question about needing this line.

PR #575, opened 2020-05-07
fix test exit code error

lamebrain:
- fix exit code error
- update mocha reporter to spec

### DIFF
--- a/docs-markdown/src/test/suite/index-coverage.ts
+++ b/docs-markdown/src/test/suite/index-coverage.ts
@@ -38,10 +38,6 @@ export function run(): Promise<void> {
 	// Create the mocha test
 	const mocha = new Mocha({
 		color: true,
-		reporter: "mocha-junit-reporter",
-		reporterOptions: {
-			mochaFile: "../../out/coverage/test-results.xml",
-		},
 		timeout: 15000,
 		ui: "tdd",
 	});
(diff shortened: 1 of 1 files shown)


docs-markdown/src/test/suite/index-coverage.ts
```diff
@@ -34,14 +34,12 @@ function setupNyc() {
...
 
 export function run(): Promise<void> {
-	let nyc = setupNyc();
+	// tslint:disable-next-line
+	let nyc = setupNyc(); // "const" causes coverage report error
+
 	// Create the mocha test
 	const mocha = new Mocha({
 		color: true,
-		reporter: "mocha-junit-reporter",
```
bharney0, line 41: I didnt notice any problems with the mocha-junit-reporter? That is how we get the reported problems with which tests had failed in CI. Do we actually want to remove this?

docs-markdown/src/test/suite/index-coverage.ts
```diff
@@ -34,14 +34,12 @@ function setupNyc() {
 }
 
 export function run(): Promise<void> {
-	let nyc = setupNyc();
+	// tslint:disable-next-line
+	let nyc = setupNyc(); // "const" causes coverage report error
```
adunndevster, line 38: This is so confusing... I have a fix for this same issue here that does use const: https://github.com/microsoft/vscode-docs-authoring/pull/574

docs-markdown/src/test/suite/index-coverage.ts
lamebrain: ahh, got it. I wanted to see the output on the console. Reverting for now. Maybe we can add both later.

PR #589, opened 2020-05-20
Add support for clickable images

jamarw:
1. Add support for clickable images
2. Fix linting errors in italics test

### DIFF
--- a/docs-markdown/src/controllers/image-controller.ts
+++ b/docs-markdown/src/controllers/image-controller.ts
@@ -3,8 +3,8 @@ import Axios from "axios";
 import { existsSync } from "fs";
 import { basename, dirname, extname, join, relative } from "path";
 import * as recursive from "recursive-readdir";
-import { CompletionItem, Position, QuickPickItem, QuickPickOptions, window, workspace } from "vscode";
-import { hasValidWorkSpaceRootPath, ignoreFiles, insertContentToEditor, isMarkdownFileCheck, isValidEditor, noActiveEditorMessage, setCursorPosition } from "../helper/common";
+import { CompletionItem, Position, QuickPickItem, QuickPickOptions, window, workspace, InputBoxOptions } from "vscode";
+import { hasValidWorkSpaceRootPath, ignoreFiles, insertContentToEditor, isMarkdownFileCheck, isValidEditor, noActiveEditorMessage, setCursorPosition, postWarning } from "../helper/common";
 import { sendTelemetryData } from "../helper/telemetry";
 
 const telemetryCommandMedia: string = "insertMedia";
@@ -54,6 +54,10 @@ export function pickImageType() {
         description: "",
         label: "Add lightbox to image",
     });
+    items.push({
+        description: "",
+        label: "Add link to image",
+    });
     window.showQuickPick(items, opts).then((selection) => {
         if (!selection) {
             return;
@@ -78,6 +82,10 @@ export function pickImageType() {
             case "add lightbox to image":
                 applyLightbox();
                 commandOption = "lightbox";
+                break;
+            case "add link to image":
+                applyLink();
+                commandOption = "link";
                 break;
         }
         sendTelemetryData(telemetryCommand, commandOption);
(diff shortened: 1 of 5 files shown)


docs-markdown/src/controllers/image-controller.ts
```diff
@@ -503,3 +512,44 @@ export function imageCompletionProvider() {
...
+
+    // if user has not selected any text, then continue
+    const RE_LINK = /:::image\s+((source|type|alt-text|lightbox|border|loc-scope)="([a-zA-Z0-9_.\/ -]+)"\s*)+:::/gm;
+    const position = new Position(editor.selection.active.line, editor.selection.active.character);
+    // get the current editor position and check if user is inside :::image::: tags
+    const wordRange = editor.document.getWordRangeAtPosition(position, RE_LINK);
+    if (wordRange) {
+        const options: InputBoxOptions = {
+            placeHolder: "Enter URL",
+            validateInput: (urlInput) => urlInput.startsWith("http://") || urlInput.startsWith("https://") ? "" :
```
bharney0, line 532: Can the urlInput aslo be a relative Url? "../../link/in/another/document.md"

docs-markdown/src/controllers/image-controller.ts
```diff
@@ -503,3 +512,44 @@ export function imageCompletionProvider() {
...
+            if (imageLink === undefined) {
+                postWarning("No URL given, abandoning command.");
+            } else {
+                editor.edit((selected) => {
+                    selected.insert(new Position(wordRange.end.line, wordRange.end.character - 3), ` link="${imageLink}"`);
+                });
+            }
+        });
+    } else {
+        const RE_LINK_EXISTS = /:::image\s+((source|type|alt-text|lightbox|border|loc-scope|link)="([a-zA-Z0-9_.\/ -:{1}]+)"\s*)+:::/gm;
```
bharney0, line 545: what is the :{1} for in this regex?

docs-markdown/src/controllers/image-controller.ts
jamarw: Hi @bharney0 , as far as I know this should only support links to web pages but I'll follow-up to make sure.

docs-markdown/src/controllers/image-controller.ts
jamarw: This is used to look for a single colon in the url to support http(s)://. The link attribute wasn't captured with the current regex because the colon didn't match the pattern.

docs-markdown/src/controllers/image-controller.ts
```diff
@@ -385,7 +394,7 @@ export async function applyLocScope() {
         return;
     }
     // if user has not selected any text, then continue
-    const RE_LOC_SCOPE = /:::image\s+((source|type|alt-text|lightbox|border)="([a-zA-Z0-9_.\/ -]+)"\s*)+:::/gm;
+    const RE_LOC_SCOPE = /:::image\s+((source|type|alt-text|lightbox|border|link)="([a-zA-Z0-9_.\/ -]+)"\s*)+:::/gm;
```
bharney0, line 397: I dont think this will support other locales/foreign languages. I think we need `.*?`. I think the same problem exists with all these regexes actually.

docs-markdown/src/controllers/image-controller.ts
jamarw: Hey @bharney0 , good call.  We should support relative paths as well.  I removed the url check.  Thanks

docs-markdown/src/controllers/image-controller.ts
jamarw: Made the suggested regex change here and I'll create a work item to change the other regex patterns so we can test for regressions.

docs-markdown/src/controllers/image-controller.ts
jamarw: https://github.com/microsoft/vscode-docs-authoring/pull/589#discussion_r428336650

PR #590, opened 2020-05-20
docs-images: Added ESLint and Prettier.

bharney0:


adunndevster:
@bharney0 - When I try to run this I get the following error when trying to compress all images in a folder:
image

Do you get the same?

PR #603, opened 2020-05-28
Improvements to the "collapse relative links in folder" experience

IEvangelist:
- Ignore changes that simply remove leading `./`
- Fix reporting
- Make cancellable
- Increment progress
- File names, not full UNC paths


NextTurn:
@IEvangelist Is it now cheaper to use a regex that matches only paths starting with `../`?

IEvangelist:
> @IEvangelist Is it now cheaper to use a regex that matches only paths starting with `../`?

That is a great question, probably!

PR #606, opened 2020-06-01
Collapse relative links: Lazily open editor

NextTurn:
@IEvangelist for review.

(I'm new to TS so please feel free to leave suggestions.)


docs-markdown/src/controllers/link-controller.ts
```diff
@@ -84,40 +84,51 @@ export async function collapseRelativeLinksInFolder(uri: Uri) {
...
 	if (!content) {
 		return;
 	}
 
-	const directory = dirname(editor.document.fileName);
-	const tempReplacements = findReplacements(editor.document, content, null, linkRegex);
+	const directory = dirname(document.fileName);
+	const tempReplacements = findReplacements(document, content, null, linkRegex);
+	if (!tempReplacements || tempReplacements.length === 0) {
+		return 0;
```
IEvangelist, line 105: Use 4x spaces, rather than tabs.

NextTurn: VS Code automatically chose the indentation, which means... ;)

bharney0: Yea, we are using the same eslint config as the docs-ui team so we are using tabs of size 2 instead of 4 spaces.

PR #609, opened 2020-06-01
Fixed Insert Link for TOC yaml and md

bharney0:
Fixed Insert Link for TOC yaml and md [AB#227286](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/227286)


docs-markdown/src/helper/common.ts
```diff
@@ -186,7 +186,6 @@ export function hasValidWorkSpaceRootPath(senderName: string) {
 
 export async function insertContentToEditor(
 	editor: vscode.TextEditor,
-	senderName: string,
```
bharney0, line 189: I removed this parameter which caused such a large diff. But the change was very small because this parameter is unused.

adunndevster: Nice clean up!

PR #610, opened 2020-06-03
Metadata controller tests

mevansgh1:


docs-markdown/src/controllers/cleanup/cleanup-controller.ts
```diff
@@ -195,10 +195,9 @@ export async function applyCleanupFolder(uri: Uri) {
 				const promises: Promise<any>[] = [];
 				// check for dirty files
 				files.map(file => {
-					const fileName = basename(file);
-					const modifiedUri = join(uri.path, fileName).replace(/\\/g, '/');
-					if (extname(modifiedUri) === '.md') {
-						workspace.openTextDocument(Uri.parse(modifiedUri)).then(doc => {
+					file = file.replace(/\\/g, '/');
```
mevansgh1, line 198: I stumbled upon what I *think* may be a bug.  When we have nested folders in the Uri we pass to  `applyCleanupFolder` the `modifiedUri` ends up with the wrong path to the file, so it never gets cleaned up. For example, in the screenshot below `/test` is stripped from the `modifiedUri`, which results in a file not found exception when running the tests 

cleanup bug

In addition to this fix I have modified one test case and added a new one.  I've modified the existing test to make sure the we can recursively cleanup folders (the issue I'm fixing).  The new test is just adding some test coverage for dirty files.  I noticed most of these tests are just expecting method calls but these new tests will verify the output which I think strengthens the test coverage.

docs-markdown/src/test/suite/controllers/cleanup/cleanup-controller.test.ts
```diff
@@ -100,6 +118,7 @@ suite('Cleanup Controller', () => {
 		expect(spy).to.have.been.called();
 	});
 	test('cleanup repo - empty metadata - empty values', async () => {
+		await loadDocumentAndGetItReady(filePath);
```
mevansgh1, line 121: In debugging I noticed that some of these tests will fail when running by themselves.  This change makes these test independent of each other.

docs-markdown/src/test/suite/controllers/include-controller.test.ts
```diff
@@ -54,15 +54,15 @@ suite('Include Controller', () => {
 	});
 	test('isMarkdownFileCheck', async () => {
 		const spy = chai.spy.on(common, 'isMarkdownFileCheck');
-		const stub = sinon.stub(glob, 'glob');
+		const stub = sinon.stub(glob, 'Glob');
```
mevansgh1, line 57: Fixes an issue I ran into when pushing my branch.

Screen Shot 2020-06-02 at 7 40 43 PM

mevansgh1:
On a side note, I had to add various fixes for eslint/compile errors after I rebased master on my branch, so this may be easier to review by looking at each commit separately.

mevansgh1:
@mamccrea

docs-markdown/src/controllers/cleanup/cleanup-controller.ts
bharney0: Sounds good, the new test looks good to me.

bharney0:
@mevansgh1 Thank you for your contributions to Docs Authoring Pack. I took a look at your changes and they look good. But I notice that I get two failing tests. Can you look into that?
image

mevansgh1:
> @mevansgh1 Thank you for your contributions to Docs Authoring Pack. I took a look at your changes and they look good. But I notice that I get two failing tests. Can you look into that?
> image

sorry about that, I added *nix line endings `\n` to the test case which works on my Mac but on Windows a carriage return was set instead `\r` so the tests failed.  I fixed this by making using `os.EOL` to make it OS agnostic.

lamebrain:
@mevansgh1 Can you take a look at this? When running the tests, I get the following error:
image

docs-markdown/src/test/suite/controllers/quick-pick-menu-controller.test.ts
```diff
@@ -1,6 +1,7 @@
 /* eslint-disable @typescript-eslint/no-non-null-assertion */
 import * as chai from 'chai';
 import * as spies from 'chai-spies';
+import * as telemetry from '../../../helper/telemetry';
```
bharney0, line 4: It looks like this is throwing an error, as the telemetry import is already happening on line 23 of this file. Please remove this.

docs-markdown/src/test/suite/controllers/table-controller.test.ts
```diff
@@ -27,6 +27,9 @@ suite('Table Controller', () => {
 	teardown(() => {
 		chai.spy.restore(common);
 	});
+	suiteSetup(() => {
```
bharney0, line 30: It looks like this suiteSetup is already being done on line 23, and is causing this test to fail. Please remove this sendTelemetryData.

docs-markdown/src/test/suite/controllers/quick-pick-menu-controller.test.ts
mevansgh1: It looks like someone beat me to the punch on this change,  another PR which was recently merged, see [here](https://github.com/microsoft/vscode-docs-authoring/commit/e372d608139a7f100390461a7c3d3d5f35d14405).  I've removed the duplicate import.  However the build check is still failing.  I've ran this locally on my Mac and on a Windows10 VM and the test pass.  Can you please share the details of the new failure?

Also there is a new check for CodeQL-Build which is failing with:


```
Run github/codeql-action/autobuild@v1
Attempting to automatically build csharp code
  /opt/hostedtoolcache/CodeQL/1.0.0/x64/codeql/csharp/tools/autobuild.sh
  Semmle autobuilder for C#
  Working directory: /home/runner/work/vscode-docs-authoring/vscode-docs-authoring
  Attempting to locate build script
  Error: Could not auto-detect a suitable build method
  ##[error]We were unable to automatically build your code. Please replace the call to the autobuild action with your custom build steps.  The process '/opt/hostedtoolcache/CodeQL/1.0.0/x64/codeql/csharp/tools/autobuild.sh' failed with exit code 1
```

I'm not familiar with this, can you please advise?

I'm not familiar with this, please advise.

docs-markdown/src/test/suite/controllers/table-controller.test.ts
mevansgh1: fixed

lamebrain:
@mevansgh1 I resolved the build errors.

PR #616, opened 2020-06-04
Add new condition to insertNestedList command

jamarw:
Docs-markdown shouldn't override `Tab` in snippets.


jamarw:
> Can we add a test for this?

Hey @lamebrain , I'll look into it.  Thanks

lamebrain:
AB#227365

Issue #490, opened 2020-03-24
Unit Test Coverage: row-columns-controller.ts


bharney0:
Add unit test coverage for row-columns-controller

PR #630, opened 2020-06-11
Added Unit test : row-columns-controller

JohnTheFarmer:
#490 @mamccrea


jamarw:
Hi @JohnTheFarmer , thanks for the contribution!  We'll start reviewing the PR soon and let you know if there's any feedback.

lamebrain:
Hi @JohnTheFarmer , when running the tests locally a bunch of them fail. The most common one is "TypeError: Attempted to wrap showQuickPick which is already wrapped". Can you take a look?

JohnTheFarmer:
Hi @lamebrain  I tested it locally , i don't have showQuickPick  error , but I do have "insertRowsAndColumnsCommand" case failed and a bunch warnings  for some tests

 "TypeError: Cannot read property 'then' of undefined"  
"TypeError: Cannot read property 'sendTelemetryEvent' of undefined" 

 for  the one failed test case , the issue seems to be in the quick-pick-menu-controller.test,  I just added one line of code to that file. Please review

PR #634, opened 2020-06-12
Adding extension.ts tests

jamarw:
[AB#233670](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/233670)


docs-markdown/src/test/suite/extension.test.ts
```diff
@@ -1,19 +1,28 @@
...
 
+suite('Extension tests', () => {
+	test('installedExtensionsCheck', () => {
+		const spy = chai.spy(installedExtensionsCheck);
 		installedExtensionsCheck();
-
-		assert.equal(-1, [1, 2, 3].indexOf(5));
-		assert.equal(-1, [1, 2, 3].indexOf(0));
+		sleep(sleepTime);
+		expect(spy).to.be.spy;
```
bharney0, line 14: Im not sure I understand what is being tested here. It looks like we are checking if the function you setup as a spy is a spy? Maybe we want to check if the function has been called and call activate?

docs-markdown/src/test/suite/extension.test.ts
```diff
@@ -1,19 +1,28 @@
...
+	});
+	test('setupAutoComplete', () => {
+		const spy = chai.spy(setupAutoComplete);
+		setupAutoComplete();
+		sleep(sleepTime);
+		expect(spy).to.be.spy;
```
bharney0, line 20: same here

docs-markdown/src/test/suite/extension.test.ts
```diff
@@ -1,19 +1,28 @@
...
+	});
+	test('deactivate', () => {
+		const spy = chai.spy(deactivate);
+		deactivate();
+		sleep(sleepTime);
+		expect(spy).to.be.spy;
```
bharney0, line 26: and here

docs-markdown/src/test/suite/extension.test.ts
jamarw: Hi @bharney0 , I'll take a look.  It should test for calls but maybe I committed this when I was testing for external extensions.

jamarw: Hi @bharney0 , forgot to push my last commit.  Can you take a look now?  Thanks!

Issue #489, opened 2020-03-24
Unit Test Coverage: preview-controller.ts


bharney0:
Add unit test coverage for preview-controller

PR #637, opened 2020-06-14
add preview controller tests

mamccrea:
#489


lamebrain:
@mamccrea Hi Mary, same with this PR, looks like there are some conflicts which are causing some test failures. Can you take a look?

docs-markdown/src/test/suite/controllers/preview-controller.test.ts
```diff
@@ -0,0 +1,58 @@
...
+	});
+	test('noActiveEditorMessage', () => {
+		const spy = chai.spy.on(common, 'noActiveEditorMessage');
+		previewTopic();
+		expect(spy).to.have.been.called();
+	});
+	test('isMarkdownFileCheck', async () => {
+		const filePath = resolve(__dirname, '../../../../../src/test/data/repo/articles/preview.md');
+		await loadDocumentAndGetItReady(filePath);
+		const stub = sinon.stub(telemetry, 'sendTelemetryData');
```
bharney0, line 36: can you put this line in a suiteSetup similar to this test https://github.com/microsoft/vscode-docs-authoring/blob/d9f16d58c84427179d24efde462ffac9658a03f4/docs-markdown/src/test/suite/controllers/image-controller.test.ts#L40. You will also need the suiteTearDown https://github.com/microsoft/vscode-docs-authoring/blob/d9f16d58c84427179d24efde462ffac9658a03f4/docs-markdown/src/test/suite/controllers/image-controller.test.ts#L35

docs-markdown/src/test/suite/controllers/preview-controller.test.ts
```diff
@@ -0,0 +1,56 @@
...
+		previewTopic();
+		expect(spy).to.have.been.called();
+	});
+	test('showPreviewToSide - checkExtension', async () => {
+		const spy = chai.spy.on(common, 'checkExtension');
+		previewTopic();
+		await sleep(100);
+		expect(spy).to.have.been.called();
+	});
+	test('seoPreview - checkExtension', async () => {
```
bharney0, line 50: This looks like the same test as above. Did you mean to do seoPreview?

mamccrea: I agree - this is leftover from when I was using the other extension in the test. I'll take it out.

mamccrea:
@lamebrain - it says it's failing on the "Common getOSPlatform returns osPlatform" test, but I'm not sure why/how my tests would impact that test.

bharney0:
> @lamebrain - it says it's failing on the "Common getOSPlatform returns osPlatform" test, but I'm not sure why/how my tests would impact that test.

Its likely that the existing call to `getOSPlatform` is interfering with the test. You may need to stub that call and restore it after the test executes.

PR #644, opened 2020-06-18
unit test for metadata-controller

lamebrain:
[AB#179029](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/179029)


docs-markdown/.vscode/launch.json
```diff
@@ -30,6 +32,22 @@
                 "${workspaceFolder}/out/src/**/*.js"
             ],
             "preLaunchTask": "mocha"
+        },
+        {
+            "name": "Metadata Controller Extension Tests",
```
jamarw, line 37: Hi @lamebrain , I think we can remove this section because mocha will pick up every test in the **test** directory by default so the existing "extension tests" section should be enough.  

I removed the "Metadata Controller Extension Tests" block then ran the tests locally and everything worked fine.

lamebrain: I added this because I didn't want to run all the tests when working on my changes. Might be useful/save time to break up the tests like this.

jamarw: Ah, makes sense.

Issue #494, opened 2020-03-24
Unit Test Coverage: template-controller.ts


bharney0:
Add unit test coverage for template-controller

mamccrea:
@JohnTheFarmer

PR #648, opened 2020-06-22
Added Unit Test : Template-controller

JohnTheFarmer:
#494


JohnTheFarmer:
@mamccrea

docs-markdown/src/test/suite/controllers/template-controller.test.ts
```diff
@@ -0,0 +1,64 @@
...
+		applyTemplate();
+		await sleep(sleepTime);
+		expect(spy).to.have.been.called();
+		stubShowQuickPick.restore();
+	});
+	test('applyTemplate - executeCommand', async () => {
+		const stubShowQuickPick = sinon.stub(window, 'showQuickPick');
+		stubShowQuickPick.onCall(0).resolves('docs-article-metadata.md');
+		const spy = chai.spy.on(commands, 'executeCommand');
+		if (common.checkExtension('docsmsft.docs-article-templates', '')) {
```
bharney0, line 57: On CI the docs-article-template isnt installed. So this will never be true when running on CI, but might work locally with the extension installed. Do we need this if statement? I think you want to have something that will test the controller without checking the extension is installed.

JohnTheFarmer: Do I remove this test case ? or remove just the if statement?

bharney0: If removing the if statement fails the test when you dont have docs-article-templates extension installed, you will need to fix the test to pass.

JohnTheFarmer:
@bharney  I don't know what happened. CI check failed first time , but passed second when i closed and reopened the pull request.I also tested it locally  after CI check failed first time, there was a error about "..stack ...adal-node.." (there was no error before the pull request) ,  i can't remember what was it exactly , but the Vscode was crushed when I ran the test few times.

PR #657, opened 2020-06-23
✨ support updating empty metadata ✨

mevansgh1:


docs-markdown/src/helper/utility.ts
```diff
@@ -530,7 +531,7 @@ export function findReplacement(
 				index += result[0].indexOf(match);
 			}
 			const startPosition = document.positionAt(index);
-			const endPosition = document.positionAt(index + match.length);
+			const endPosition = new Position(startPosition.line, startPosition.character + match.length);
```
mevansgh1, line 534: I'm working on a different PR and I stumbled upon an odd behavior I think can be improved.  The use case I had was updating the `ms.date` metadata field with an empty date.  I noticed that the line after this field was deleted.  For example in the image below see how `ms.author` gets removed

ezgif com-video-to-gif

I've added a fix for this and a test case along with it.  The new behavior is shown below, where `ms.author` remains
ezgif com-video-to-gif(1)


**Note** - This fix only applies for the `findReplacement()` method.  I did notice this same line of code [here](https://github.com/microsoft/vscode-docs-authoring/blob/master/docs-markdown/src/helper/utility.ts#L502) in the `findReplacements()` (plural) method.  I didn't spend anytime though investigating whether or not this fix would be appropriate for that workflow but I imagine so, in case anyone wants to weigh-in?

PR #658, opened 2020-06-23
Update Mocha to latest version

jamarw:


bharney0:
@jamarw were the @types upgraded at all? It might be worth seeing if we can get latest on that too.

jamarw:
> @jamarw were the mocha types upgraded at all? It might be worth seeing if we can get latest on that too.

Good call, small change but we should still have the latest.

Issue #361, opened 2019-12-02
add metadata nag options


meganbradley:
1. I edit my topic in VS Code. 
2. I save my topic. 
3. Upon save, the Docs Markdown extension checks whether I have changed the value of ms.date. If not, I get a banner: "You have changed this file. Do you want to update ms.date with today's date?" 
   - If I select yes, ms.date is set to today's date in the correct format. 
   - If I select no, no change is made, and I am not asked again for this file in this VS Code session. 

This setting is enabled by default, but can be turned off in VS Code settings, e.g. "nag to update date = false".

Similarly, the extension checks the author and ms.author values against the values set in my settings (if applicable), and if they're different asks if I want to update them to my values.

Challenge: How to handle auto-save. Ask after file has been open for 2 minutes, and if dismissed don't ask again?

nschonni:
An alternate/additional approach would be to have a third `auto` option, that would auto-update the date on save. I wouldn't suggest that as a default, and maybe you already discounted that as part of your internal discussions.

IEvangelist:
I believe that we should rely on the `docfx.json` file for the `author`, `ms.author`, `ms.service`, and `ms.subservice` details as they will commonly exist as a glob pattern to desired default in docs repos. They are specified under the `fileMetadata` node - I'm trying to find time to work on this one.

mamccrea:
We have the nag working for `ms.date`. Do we want nag options for the `author` values in the interim?

IEvangelist:
> We have the nag working for `ms.date`. Do we want nag options for the `author` values in the interim?

Hi @mamccrea,

What does the nag do, could you share a screenshot? Curious about where it shows up and how it is actionable. I would imagine that content developers would appreciate the gentle reminder about the `ms.date` but not about the `author` -- that seems less likely to change.

mamccrea:
@IEvangelist - I agree with your thoughts about an author nag. I’ll share a screenshot, too, but on save, the command palette drops down if the date is not today’s date and asks the user if they want to update the date. Enter for yes, esc for no.

IEvangelist:
We already have tooling in place that could be used to update the `ms.date`, are you familiar with that?

https://docs.microsoft.com/en-us/contribute/docs-authoring/metadata-updates

mamccrea:
Yes, I am familiar. The difference is that the nag option prompts the author on save instead of the author needing to remember to update the date, which was the request in this issue.

bharney0:
[...]

PR #669, opened 2020-06-30
Msdate nag

mevansgh1:
Issue #361 

nag msdate


docs-markdown/src/helper/utility.ts
```diff
@@ -525,7 +527,7 @@ export function findReplacement(
 	if (result !== null && result.length) {
 		const groups = !isRegExp(expression) ? expression.groups : null;
 		const match = groups && result.groups ? result.groups[groups[0]] : result[0];
-		if (match) {
+		if (match && match !== value) {
```
mevansgh1, line 530: This is unrelated to my changes, just something I noticed.  There's no need to return a replacement if it's already the value we want to set.  This saves some branches of code from executing when someone attempts to update the ms.date when it's already the current date.  Just a nit pick fix

docs-markdown/src/helper/utility.ts
```diff
@@ -442,7 +442,9 @@ export async function findAndReplaceTargetExpressions(event: TextDocumentChangeE
 					}
 				);
 
-				await applyReplacements(replacements, editor);
+				if (replacements.length > 0) {
```
mevansgh1, line 445: No need to call apply if there is nothing to apply

docs-markdown/src/test/suite/controllers/quick-pick-menu-controller.test.ts
```diff
@@ -73,6 +73,14 @@ const context: ExtensionContext = {
 suite('Quick Pick Menu Controller', () => {
 	suiteSetup(() => {
 		sinon.stub(telemetry, 'sendTelemetryData');
+		sinon.stub(workspace, 'getConfiguration').returns({
```
mevansgh1, line 76: There is a convention throughout the unit tests to override methods such as `getConfiguration()`.  However, this can result in unexpected behavior if someone is writing a new test and expecting the default behavior.  The use case I had was writing a new test case and expecting the configurations to be returned.  These overrides prevented my test from receiving the valid configuration.

For example, I have added config property 

```
{
  "markdown.metadataNag": true
}
```
to path `docs-markdown/src/test/data/repo/.vscode/settings.json`.  When I tried to load this using `workspace.getConfiguration()` it returns undefined
Screen Shot 2020-06-29 at 5 41 52 PM

By converting these overrides to stub we get the default behavior in new tests where it is defined.   You can see below how the property returns. This also allows us to not repeat ourselves with code in the unit tests.

Screen Shot 2020-06-29 at 5 51 58 PM

docs-markdown/src/test/suite/controllers/image-controller.test.ts
```diff
@@ -44,17 +44,7 @@ const context: ExtensionContext = {
 	storagePath: '',
 	globalStoragePath: '',
 	logPath: '',
-	extensionUri: Uri.parse(uri),
-	environmentVariableCollection: {
```
mevansgh1, line 48: Fixes

Screen Shot 2020-06-29 at 6 37 00 PM

mevansgh1:
I'm not sure why the build is failing it passes for me locally on both a Mac Os and Windows 10 VM.  Can someone please share the details of the build failure?

docs-markdown/src/helper/metadata.ts
```diff
@@ -0,0 +1,48 @@
...
+import { findReplacement } from './utility';
+
+const blacklist: string[] = [];
+
+export async function nag() {
+	if (workspace.getConfiguration('markdown').metadataNag === false) {
+		return;
+	}
+	const editor = window.activeTextEditor;
+	if (!editor) {
```
nschonni, line 13: Is this catching the case when doing a large Find/Replace?

mevansgh1: Sorry I'm not sure I understand your question.   Can you please clarify? The line of code you commented on simply returns if there is no active editor.

This function is triggered from an onsave event so if someone were to do a large find/replace it would only trigger the function once IF auto-saved were enabled, assuming the find/replace is an atomic transaction.  

If auto save is not enabled then find/replace should not matter since this is only triggered when save is called.

lamebrain:
@mevansgh1 Here's the error from the build:

Unable to write to Workspace Settings because markdown.metadataNag is not a registered configuration.

Error: Unable to write to Workspace Settings because markdown.metadataNag is not a registered configuration.
    at e.reject (file:///D:/a/1/s/docs-markdown/.vscode-test/vscode-1.46.1/resources/app/out/vs/workbench/workbench.desktop.main.js:5363:839)
    at e.resolveAndValidate (file:///D:/a/1/s/docs-markdown/.vscode-test/vscode-1.46.1/resources/app/out/vs/workbench/workbench.desktop.main.js:5366:717)
    at e.doWriteConfiguration (file:///D:/a/1/s/docs-markdown/.vscode-test/vscode-1.46.1/resources/app/out/vs/workbench/workbench.desktop.main.js:5361:70)
    at Object.factory (file:///D:/a/1/s/docs-markdown/.vscode-test/vscode-1.46.1/resources/app/out/vs/workbench/workbench.desktop.main.js:5360:801)
    at u.consume (file:///D:/a/1/s/docs-markdown/.vscode-test/vscode-1.46.1/resources/app/out/vs/workbench/workbench.desktop.main.js:130:75)
    at file:///D:/a/1/s/docs-markdown/.vscode-test/vscode-1.46.1/resources/app/out/vs/workbench/workbench.desktop.main.js:129:886
    at new Promise (<anonymous>)
    at u.queue (file:///D:/a/1/s/docs-markdown/.vscode-test/vscode-1.46.1/resources/app/out/vs/workbench/workbench.desktop.main.js:129:810)
    at e.writeConfiguration (file:///D:/a/1/s/docs-markdown/.vscode-test/vscode-1.46.1/resources/app/out/vs/workbench/workbench.desktop.main.js:5360:786)
    at C.writeConfigurationValue (file:///D:/a/1/s/docs-markdown/.vscode-test/vscode-1.46.1/resources/app/out/vs/workbench/workbench.desktop.main.js:5389:83)
    at file:///D:/a/1/s/docs-markdown/.vscode-test/vscode-1.46.1/resources/app/out/vs/workbench/workbench.desktop.main.js:5377:320
    at processTicksAndRejections (internal/process/task_queues.js:85:5)

docs-markdown/src/helper/metadata.ts
nschonni: I was wondering if this section of code was protecting against the nag running on a large Find/Replace when run in VS Code. As a fake example, if you do a find for the letter `a`, and replace it with `b`, are you going to have 100+ nags stacked on each other?

docs-markdown/src/test/suite/controllers/image-controller.test.ts
mevansgh1: This error actually was related to me having a newer version of the `@types/vscode` package because my package-lock.json had a different version in it then what was in package.json.  

In this repository I see 8 extensions, 6 of which commit their package-lock.json.  Only the *docs-images* and *docs-markdown* extensions do **not** commit their package-lock.json.

per the documentation [configuring-npm](https://docs.npmjs.com/configuring-npm/package-lock-json.html)

`This file is intended to be committed into source repositories, and serves various purposes`

I have added a new commit to this PR to commit this file.  I'm not sure why this was originally .gitignored, but please review this comment.

docs-markdown/src/helper/metadata.ts
mevansgh1: I tested this locally with auto-save enabled and vscode only triggers 1 onSave event when running a find/replace on the same string. I did add code below to make sure we only nag once per file in a session.  So this code will also prevent the issue you're describing IF vscode did trigger 100 onsave events during a find/replace.

```
// skip files that have already been saved
	if (blacklist.indexOf(fileName) > -1 === true) {
		return;
	}
	blacklist.push(fileName);
```

docs-markdown/src/test/suite/controllers/image-controller.test.ts
mevansgh1: disregard will submit in separate PR

jamarw:
Hi @mevansgh1 , just wanted to let you know we're reviewing this PR now.  Thanks!

docs-markdown/src/helper/metadata.ts
```diff
@@ -0,0 +1,48 @@
...
+		return;
+	}
+
+	if (!isMarkdownFileCheck(editor, false)) {
+		return;
+	}
+
+	const fileName = editor.document.fileName;
+	// skip files that have already been saved
+	if (blacklist.indexOf(fileName) > -1 === true) {
```
IEvangelist, line 24: Please reconsider naming this variable to something other than `blacklist`, perhaps `disallowlist`?

docs-markdown/src/helper/metadata.ts
```diff
@@ -0,0 +1,48 @@
...
+		const msDateRegex = /ms.date:\s*\b(.+?)$/im;
+		const msDateReplacement = findReplacement(
+			editor.document,
+			content,
+			`ms.date: ${toShortDate(new Date())}`,
+			msDateRegex
+		);
+		if (msDateReplacement) {
+			const syncDate = await window.showInformationMessage(
+				"Would you like to update ms.date to today's date?",
```
IEvangelist, line 40:
```suggestion
				"Would you like to update `ms.date` to today's date?",
```

I believe that we can format this as code, and it will render as such in the prompt.

docs-markdown/src/helper/metadata.ts
jamarw: Hi @IEvangelist , good call. I believe there's a feature/work item to rename all of these instances soon.  Thanks

PR #670, opened 2020-07-01
Auto-Complete for front-matter metadata

bharney0:
Added metadata completion provider for auto-complete. Adding auto-complete on metadata front-matter values for ms.prod, ms.service, ms.technology, and ms.subservice.


bharney0:
> tests?

Added unit tests for the completion providers

PR #675, opened 2020-07-08
Cleanup script for bad devlangs in code blocks

jamarw:
[AB#235780](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/235780)


docs-markdown/src/test/suite/controllers/cleanup/devlangsInCodeBlocks.test.ts
```diff
@@ -0,0 +1,23 @@
...
+		findCodeBlocks('```json');
+		await sleep(sleepTime);
+		expect(spy).to.have.been.called;
+	});
+	test('lowerCaseDevlang', async () => {
+		const data = '```JSON';
+		const output = convertDevlang(data, regex);
+		expect(output).to.equal('```json');
+	});
+});
```
lamebrain, line 23: Can we also add tests for cs > csharp, markdown > md ?

PR #694, opened 2020-07-29
Hackathon project - Insert / update Jupyter notebook from GitHub

IEvangelist:
# Show Me the Notebook | Hackathon 2020

Authors: [Sheri Gilley](@sdgilley) & [David Pine](@IEvangelist)

Insert the markdown version of a Jupyter notebook from GitHub into a document.  When the notebook content changes, you can quickly update that markdown.

There are two commands added to the command palette:

* **Docs: Insert Jupyter notebook** - enter the URL of the notebook.  A markdown version of the notebook is added to the document starting at the position of your cursor.  Do not modify the start or end tags; it's what the next function uses to update the notebook.  

* **Docs: Update Jupyter notebook** in the start tag, re-read the notebook and replace the contents between start and end with the latest version.  If you insert multiple notebooks in the same document (why would you do that?) only the first one will be updated.

showmeInsert

Enjoy!

### DIFF
--- a/docs-markdown/src/controllers/notebook-controller.ts
+++ b/docs-markdown/src/controllers/notebook-controller.ts
@@ -1,0 +1,207 @@
+import * as os from 'os';
+import fetch from 'axios';
+import {
+	insertContentToEditor,
+	findLineNumberOfPattern,
+	noActiveEditorMessage
+} from '../helper/common';
+import { resolve, join } from 'path';
+import { Selection, window, TextEditor } from 'vscode';
+import { spawn, execSync } from 'child_process';
+import { unlinkSync, writeFileSync } from 'fs';
+import { URL } from 'url';
+import { Command } from '../Command';
+
+export const notebookControllerCommands: Command[] = [
+	{ command: insertNotebook.name, callback: insertNotebook },
+	{ command: updateNotebook.name, callback: updateNotebook }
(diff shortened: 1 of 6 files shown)


docs-markdown/src/controllers/notebook-controller.ts
```diff
@@ -0,0 +1,207 @@
...
+
+async function getRawNotebookJson(url: string): Promise<string> {
+	const response = await fetch(url);
+	const data = await response.data;
+
+	return JSON.stringify(data);
+}
+
+function wrapMarkdownInTemplate(content: string, url: string) {
+	// Advance all the headings by one, avoid multiple H1s
```
sdgilley, line 110: I think we should remove this, and leave it to the user to decide how to handle it if the document has multiple h1's.

docs-markdown/src/controllers/notebook-controller.ts
```diff
@@ -0,0 +1,207 @@
...
+	content = content.replace(/(^|\r|\n|\r\n)#/g, '\n##');
```
IEvangelist, line 111:
```suggestion
```

bharney0:
@IEvangelist this is awesome! Nice work. Ill take a look at the code and make suggestions if needed.

PR #706, opened 2020-08-03
Remove the file I/O from the jupyter notebook, making it more cross-plat

IEvangelist:
Remove file I/O of Jupyter notebook, and instead leverage the JSON in memory - passing it as `stdin` to nbcovert.

/cc @sdgilley


IEvangelist:
I'm still having issues with getting this to function correctly on macOS. There is a `stdin` character limit of 1024. Trying to figure that out.

PR #709, opened 2020-08-05
Remove output channel from common

jamarw:
Docs-preview is creating multiple output channels.  We only need the channel created in extension.ts so this PR removes the additional channel spun up in common.ts.

image


lamebrain:
@jamarw can you check why there is such a big coverage drop?

jamarw:
> @jamarw can you check why there is such a big coverage drop?

Will do

Issue #492, opened 2020-03-24
Unit Test Coverage: sort-controller.ts


bharney0:
Add unit test coverage for sort-controller

mamccrea:
@rsknutson

PR #737, opened 2020-09-05
Sort controller Unit Test Coverage #492

rsknutson:
Unit test for sort-controller


bharney0:
@rsknutson we have changed the root directory to contain a `packages` folder where all of the extensions live. Can make sure the changes are made in `packages/docs-markdown` folder? So your new path would be `packages/docs-markdown/src/test/suite/controllers/sort-controller.test.ts`

rsknutson:
Opened new PR

PR #757, opened 2020-09-27
Bundle docs validation to the pack

928PJY:
[AB#288438](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/288438)


928PJY:
One question here, why do you use both `extensionPack` and `extensionDependencies `, IMO, the `extensionPack` is enough for your purpose, here are the words copied from [VS Code document](https://code.visualstudio.com/api/references/extension-manifest):

> An Extension Pack should not have any functional dependencies with its bundled extensions and the bundled extensions should be manageable independent of the pack. If an extension has a dependency on another extension, that dependency should be declared with the extensionDependencies attribute

nschonni:
Don't know if the submodule is needed

herohua (approved):

928PJY:
@nschonni You can reference the design spec in the linked Azure DevOps task, it is already reviewed with @bharney and @lamebrain .

nschonni:
@928PJY can't actually, since the DevOps instance is private, and I'm not a Microsoft employee 😉

bharney0 (approved):

bharney0:
@nschonni going forward we would like to use submodules to hopefully get around and issue we found with Husky. We noticed we could not control the Husky pre-push hooks where the git directory was in the root with packages in subdirectories. So we added learna and are now running all of the linting on all extensions when pushing. However, it would be nice just to have one run on the specific extension when committing changes.

PR #765, opened 2020-10-11
Added New Feature : YAML Landing Preview

JohnTheFarmer:
@mamccrea  @bryce-turner


JohnTheFarmer:
update-change (1)
syntax-landing
error (2)

JohnTheFarmer:
syntax-landing

JohnTheFarmer:
error (2)

JohnTheFarmer:
changeTheme

packages/docs-preview/src/extension.ts
```diff
@@ -116,15 +150,54 @@ export async function activate(context: ExtensionContext) {
...
+					yamlPanel = undefined;
+				},
+				null,
+				context.subscriptions
+			);
+		};
+	}
+
+	async function getStyleHtml(uri: Uri) {
+		const stylePath = Uri.file(join(context.extensionPath, 'media', 'yamlPreview.css'));
```
bharney0, line 192: I would recommend you include https://github.com/microsoft/vscode-docs-authoring/blob/master/packages/docs-preview/media/site-ltr.css which is the docs styles.

JohnTheFarmer: @bharney0 I tried use site-ltr.css at first , but I feel the result looks a little weird in side view. the icon seems smaller (on both side and full view) , and missing "docon-topic-video "...   yamlPreview.css is modified / simplified version of site-ltr , full view in side view

Testing content: https://docs.microsoft.com/en-us/azure/stream-analytics/

### half screen modified vs original


### full screen modified vs original

  

#### let me know if you still want  site-ltr.css instead of yamlPreview.css or any further changes are needed.

bharney0:
@meganbradley can you take a look at this feature to see what else might need to happen for this?

packages/docs-preview/src/extension.ts
bharney0: You can keep the smaller file if it renders a better preview.

bharney0:
@JohnTheFarmer can you make sure this preview only opens/displays for yaml files that start with `### YamlMime:Landing` We dont want to show this option if the user is displaying a TOC or Learn index.yml

bharney0:
@JohnTheFarmer I also think the font sizes differ from the published version. Can you try to fix the font sizes to match up with the live version? You can see the difference here:
image

packages/docs-markdown/src/controllers/quick-pick-menu-controller.ts
```diff
@@ -173,6 +173,10 @@ export function markdownQuickPick(context: ExtensionContext) {
 		{
 			description: '',
 			label: '$(lock) Non-localizable text'
+		},
+		{
+			description: '',
+			label: '$(browser) Yaml preview'
```
bharney0, line 179: When I navigate to a yaml or markdown file I dont see this option. Can you check that this is working correctly? We only want to show this Yaml preview when the user is in a .yml file.

JohnTheFarmer:
Hi @bharney0 for the font issue , did you test it with 2x zoom ? I think that might be the cause of different font size

JohnTheFarmer:
> @JohnTheFarmer can you make sure this preview only opens/displays for yaml files that start with `### YamlMime:Landing` We dont want to show this option if the user is displaying a TOC or Learn index.yml. It doesnt look like it renders `### YamlMime:Hub` type pages. That might be nice to include those seeing as they are such similar page types. However, if you dont want to take that on just make sure it doesnt appear as an option without the correct page type.

 so On non-landing/Hub YAML file  , the command 'yamlPreview' will still show but just don't display  webview panel ? do i keep the warning message "YAML Preview currently only supports Microsoft YamlMime:Landing" ?

JohnTheFarmer:
> @JohnTheFarmer can you make sure this preview only opens/displays for yaml files that start with `### YamlMime:Landing` We dont want to show this option if the user is displaying a TOC or Learn index.yml. It doesnt look like it renders `### YamlMime:Hub` type pages. That might be nice to include those seeing as they are such similar page types. However, if you dont want to take that on just make sure it doesnt appear as an option without the correct page type.

Do I keep the warning message ?

packages/docs-markdown/src/controllers/quick-pick-menu-controller.ts
JohnTheFarmer: I just tested , its showing on my end , not sure what cause it.

bharney0:
@JohnTheFarmer After reseting the zoom, I think its fine with the current font-size. And it looks like there is a message for other pages that dont have the `### YamlMime:Landing` so I think we are all good. I dont have any other feedback, now we are just waiting on the PM @meganbradley and @lamebrain for any further requirements. But I think we are good to go.

bharney0 (approved):

JohnTheFarmer:
@bharney0 I am still working on ### YamlMime:Hub , should be able to finish it today or tomorrow

PR #771, opened 2020-10-25
Added New Feature : YAML Preview for YamlMime: Hub | YamlMime:Landing

JohnTheFarmer:
@mamccrea  @bryce-turner


lamebrain:
Hi @JohnTheFarmer, first of all, thanks for your many contributions! After review and careful consideration, I do not think that this feature is a long term fit for the Docs Authoring Pack. Although I agree this brings value to the authors, the implementation is not maintainable long term as markup changes will need to be manually updated and kept in sync. Furthermore, we have plans on the roadmap to bring in a local preview based on the content build system that will not require any change maintenance. 

Like I said, I do believe that this brings value to authors and a possible path forward is to roll this out into a separate extension with a less expectations for long term support.

Thanks again for your contributions, let me know if you have any questions.

JohnTheFarmer:
@lamebrain I agree , I was worried about the maintainability. a lot changes will be needed if the live website changes ,  I tried to improve the process by refactoring my code , so others can understand and update the changes easier. … I thought of creating a program to generate html with  YAML file and website source code as inputs, but not sure how to start it. 

Anyway , I learned new things , it was a good practice :smile:

lamebrain:
@JohnTheFarmer Thanks for understanding. In the future, if you have an idea for a sizeable feature, feel free to do a write up on it first so that we can discuss the feasibility and/or product roadmap impact.

PR #773, opened 2020-11-01
add commands to enable and disable the metadata msdate reminder

mevansgh1:
@mamccrea


packages/docs-markdown/src/helper/metadata.ts
```diff
@@ -20,6 +23,11 @@ export async function nag() {
 		return;
 	}
 
+	const config = workspace.getConfiguration('markdown');
+	if (!config.get<boolean>('enableMetadataDateReminder')) {
+		return;
+	}
+
```
lamebrain, line 30: I know this isn't the correct spot in the code, but can we also suppress the reminder if the date is already the current date?

mevansgh1: There is already a guard in place for that on [line #47](https://github.com/microsoft/vscode-docs-authoring/blob/5e0ccc1ee5eec9c2ed5adf01e499f027f42d73e0/packages/docs-markdown/src/helper/metadata.ts#L47).  I had modified the `findReplacement()` function in my [initial implementation](https://github.com/microsoft/vscode-docs-authoring/commit/7668863507f8b62f60334dbb056575130054e8f6#diff-9ed3a0dd3f5cf64cd8aa41d910ff3c7d936705743a967a37db234c540e7a5e93R532) to only return a match if it was different then the initial value, otherwise it's undefined.  In this case we pass the current date to `findReplacement()` to compare against `ms.date`.  The function will return undefined if they match (i.e the ms.date is the current date) and the conditional guard on line 47 will skip the prompt to update the date.

mevansgh1: @lamebrain Can you please review this when you have a moment?

jamarw:
"/azp run docs-markdown CI"

bharney0 (approved):

bharney0:
@mevansgh1 I tried to deploy this but I am getting an error on the unit tests with this message:  `Unable to write to Workspace Settings because markdown.enableMetadataDateReminder is not a registered configuration.` would you be able to see if you can replicate and address the issue?
image

PR #819, opened 2021-02-24
Fixed devlang regex for lowercasing and fixing language paramters

bharney0:
- Fixed devlang regex for lowercasing and fixing language paramters
AB#381113


nschonni:
Is it worth supporting both ~ and ` as the code fence characters?

bharney0:
@nschonni, I think we were just going to scope the code fence characters to `. But maybe in the future we can include ~

lamebrain (approved):

PR #881, opened 2021-06-02
Fix heading command

jamarw:
[AB#435572](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/435572)

### DIFF
--- a/packages/docs-markdown/src/controllers/yaml/showTOCQuickPick.ts
+++ b/packages/docs-markdown/src/controllers/yaml/showTOCQuickPick.ts
@@ -1,20 +1,30 @@
 import { createEntry } from './createEntry';
 import { readFileSync } from 'fs';
-import { basename, dirname, join, relative } from 'path';
+import * as recursive from 'recursive-readdir';
+import { basename, dirname, extname, join, relative } from 'path';
 import { QuickPickItem, window, workspace } from 'vscode';
-import { noActiveEditorMessage, postError, showStatusMessage } from '../../helper/common';
+import { ignoreFiles, noActiveEditorMessage } from '../../helper/common';
 import { noHeadingSelected } from '../../constants/log-messages';
 import { getHeadings } from '../../helper/getHeader';
-import fg = require('fast-glob');
 
 export async function showTOCQuickPick(options: boolean) {
+	const markdownExtensionFilter = ['.md'];
 	let folderPath: string = '';
+	let fullPath: string = '';
 
 	if (workspace.workspaceFolders) {
 		folderPath = workspace.workspaceFolders[0].uri.fsPath;
 	}
 
-	const items = getQuickPickItems(folderPath);
+	const files = await recursive(folderPath, ignoreFiles);
+
+	const items: QuickPickItem[] = [];
+	files.sort();
+	files
+		.filter((file: any) => markdownExtensionFilter.indexOf(extname(file.toLowerCase())) !== -1)
+		.forEach((file: any) => {
+			items.push({ label: basename(file), description: dirname(file) });
+		});
 
 	// show the quick pick menu
 	const selectionPick = await window.showQuickPick(items);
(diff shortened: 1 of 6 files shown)


packages/docs-markdown/src/controllers/yaml/showTOCQuickPick.ts
```diff
@@ -1,20 +1,30 @@
...
 	}
 
-	const items = getQuickPickItems(folderPath);
+	const files = await recursive(folderPath, ignoreFiles);
+
+	const items: QuickPickItem[] = [];
+	files.sort();
+	files
+		.filter((file: any) => markdownExtensionFilter.indexOf(extname(file.toLowerCase())) !== -1)
+		.forEach((file: any) => {
```
IEvangelist, line 25:
```suggestion
		.filter((file: string) => markdownExtensionFilter.indexOf(extname(file.toLowerCase())) !== -1)
		.forEach((file: string) => {
```

If these are of `string` type, it should be explicit - saying they're `any` is rather misleading.

packages/docs-markdown/src/controllers/yaml/showTOCQuickPick.ts
```diff
@@ -29,31 +39,33 @@ export async function showTOCQuickPick(options: boolean) {
...
 
-	if (!title) {
-		postError(`Could not find H1: ${fullPath}`);
-		showStatusMessage(`Could not find H1: ${fullPath}`);
+	if (!headings) {
+		window.showErrorMessage(headings[0]);
 		return;
 	}
-
+	let headingName = headings.toString().replace('# ', '').split(',')[0];
```
IEvangelist, line 53:
```suggestion
	let headingName = headings.replace('# ', '').split(',')[0];
```

The `headings` variable should already be of type `string` according to [this](https://github.com/microsoft/vscode-docs-authoring/blob/74155c4528c8ce9ff19d8fa9a9dec7448c32be79/packages/docs-markdown/src/helper/getHeader.ts), no reason to call `.toString()`

packages/docs-markdown/src/test/suite/helper/getHeader.test.ts
```diff
@@ -7,12 +7,9 @@ const expect = chai.expect;
...
 	test('Get heading', async () => {
-		const filePath = resolve(
-			__dirname,
-			'../../../../../src/test/data/repo/articles/docs-markdown.md'
-		);
+		const filePath = resolve(__dirname, '../../../../../src/test/data/repo/articles/bookmark.md');
 		const content = readFileSync(filePath, 'utf8');
 		const heading = getHeadings(content);
-		expect(heading).to.equal('Docs Markdown Reference');
+		expect(heading[0]).to.equal('# This is a bookmark page');
```
IEvangelist, line 13: Could we had more bookmark tests, to ensure that multiple H{n} values work? Also, I thought that hand made bookmarks used to work too?

For example:

```md
<a name="custom-bookmark-name"></a>
```

It would find "custom-bookmark-name" too, is that not the case anymore?

packages/docs-markdown/src/controllers/yaml/showTOCQuickPick.ts
jamarw: Hey @IEvangelist , thanks for the suggestions! I committed the previous suggestion but this one causes an issue because the headings are objects and not strings at this point. Here's the console output:

image

packages/docs-markdown/src/test/suite/helper/getHeader.test.ts
bharney0: Seems like that would be good to find custom bookmarks, it wouldnt be too hard to add this back if you just made a function to find all anchor tags with name attribute.

jamarw: I looked into the custom/handmade bookmark link and it doesn't look like we supported that type of link. I tested against a few of the earlier versions of the extension and none of them supported it. 

We can definitely roll in a few more tests.

jamarw: Created custom bookmark [task ](https://ceapex.visualstudio.com/Engineering/_boards/board/t/AXE%20Team/Stories/?workitem=435760)for tracking and requirements.

bharney0 (approved):

PR #890, opened 2021-06-21
Refresh metadata explorer if editor is closed

gewarren:


gewarren:
@IEvangelist This shows the welcome message in the Metadata Explorer if the last open Markdown file is closed.

packages/docs-markdown/src/extension.ts
```diff
@@ -201,10 +201,8 @@ export async function activate(context: ExtensionContext) {
 			}
 		}
 	});
-	window.onDidChangeActiveTextEditor(e => {
-		if (['markdown', 'json'].includes(e.document.languageId)) {
-			metadataTreeProvider.refresh();
-		}
+	workspace.onDidCloseTextDocument(e => {
+		metadataTreeProvider.refresh();
```
gewarren, line 205: When the explorer refreshes, it checks the language, so we don't need to check here. Also, we want to update the explorer even if the active document isn't Markdown (i.e. show the welcome message).

IEvangelist (approved):

IEvangelist:
Attention: @bharney0

bharney0 (approved):

PR #892, opened 2021-06-24
Metadata explorer click

IEvangelist:
Added the ability for the entries in the Metadata tree view to be clickable:

- Deleted "implicit" updates
- Major clean up
- Encapsulated a bunch of bits
- Make the traverse option a fallback if glob doesn't work
- Delete *\*-refresh.svg* images, as we can use codeicon instead

/cc: @gewarren


packages/docs-markdown/src/extension.ts
```diff
@@ -201,8 +201,10 @@ export async function activate(context: ExtensionContext) {
 			}
 		}
 	});
-	workspace.onDidCloseTextDocument(e => {
-		metadataTreeProvider.refresh();
+	workspace.onDidChangeTextDocument(e => {
+		if (e.document.languageId === 'markdown') {
```
gewarren, line 205: Why did you change this? We want it to refresh and show the welcome text if the user closes the last Markdown file. Also if the user switches active focus to a non-Markdown file, we also want to refresh so we clear the tree (since it doesn't apply to the current editor).

packages/docs-markdown/src/controllers/metadata/metadata-tree-node.ts
```diff
@@ -26,8 +36,27 @@ export class MetadataTreeNode extends TreeItem {
 		this.key = entry.key;
 		this.source = entry.source;
 		this.value = entry.value;
+		// this.resourceUri = entry.resourceUri;
```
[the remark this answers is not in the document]
IEvangelist:
```suggestion

```

packages/docs-markdown/src/extension.ts
```diff
@@ -201,8 +201,10 @@ export async function activate(context: ExtensionContext) {
...
+			metadataTreeProvider.refresh();
+		}
```
IEvangelist, line 207:
```suggestion
		metadataTreeProvider.refresh();
```

packages/docs-markdown/src/extension.ts
IEvangelist: I believe that this still works the way we want it, I do see the welcome page. I'll remove the guard for `'markdown'` though.

IEvangelist:
Hi again @bharney0 - I'm fixing the bug in the unit test, as discovered from this test run failure. Should have it soon.

bharney0 (approved):

PR #899, opened 2021-07-12
Remove unused file-system dependency from docs-metadata

jamarw:
Remove unused file-system dependency to fix security issue with utils-extend

image


jamarw:
AB#447931

stevenkuhn (approved):
:+1:

